PR #7, opened 2022-06-25
Update README.md

bentobox19:
Some people may have concerns over a `curl | bash` proposition.

Also it's a good way to invite the user to find out what is Foundry.


SunXrex:
Good suggestion! Thanks @bentobox19

SunWeb3Sec:
Good suggestion! Thanks @bentobox19

PR #25, opened 2022-08-17
Add demo gif of producing gas report for the audius poc

m0ham3dx:
- I have added a demo gif of how to use `forge test --gas-report` which also pretty prints a gas usage table per function call. 
- I have also add a short paragraph explaining it
-


SunWeb3Sec:
Awesome! Thanks a lot!

PR #69, opened 2022-10-11
Fix XaveFinance PoC

DaramG:
To call `daoModule.executeProposalWithIndex`, it requires submiting some answer for that as the attacker did.


SunWeb3Sec:
Awesome! Thanks for submitting the PR.

PR #154, opened 2023-01-08
fix Harmony_multisig

rriescog:
Fix by:

- Updating block number before real hack
- Fix calldata to take address(this) to be used as recipient of the hack (balance was wrong not showing any USDT so I realized that something was not ok)
- Remove hardcoded transaction id taking it from the multisig return uint.

Result:

Captura de Pantalla 2023-01-08 a las 23 14 37


SunWeb3Sec:
Thanks mate!

PR #316, opened 2023-05-31
Add FAPEN and NOON example

Kkaminsk86:


Kkaminsk86:
Hi, I have added NOON Token example in the meantime

PR #388, opened 2023-08-09
fomatting: format files with forge fmt

akshaynexus:


SunWeb3Sec:
thanks @akshaynexus 
I will review it once I have more time.

PR #468, opened 2023-11-10
Fixed broken README links of past incidents

eloi010:
From my understanding, my proposed fix would improve the linking of past incidents and link sharing of past incidents.

The current solution only works (at least for me) if I have past years unfolded (bottom of the document). Furthermore, this could allow you to reduce the main `Readme.md` file by deleting all the past incident info from it.


SunWeb3Sec:
Thanks mate, make sense!

PR #573, opened 2024-03-06
2023-11-22 Kyberswap elastic hack

apehex:
Hey there, here's a draft for the hack of Kyberswap elastic at the end of 2023.

It is the first exploit of the chain of pool draining, in transaction [0x485e08dc2b6a4b3aeadcb89c3d18a37666dc7d9424961a2091d6b3696792f0f3](https://phalcon.blocksec.com/explorer/tx/eth/0x485e08dc2b6a4b3aeadcb89c3d18a37666dc7d9424961a2091d6b3696792f0f3?line=29).

There were actually dozens of pool drainings, following the same pattern.
Is it ok to include only this particular hack?

Also the hack is quite sophisticated (for me >>): it relies on precise calculations / values.
For now I have mostly hard-coded those from the traces.

It doesn't show how the hacker came up with those specific values though, should I go further and reverse the process?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 **Reproduce DeFi hack incidents using Foundry.**
 
-364 incidents included.
+365 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ All articles are also published on [Substack](https://defihacklabs.substack.com/
 
 [20231125 TheNFTV2](#20231125-thenftv2---logic-flaw)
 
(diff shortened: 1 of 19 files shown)


F01ivor4:
I have reversed the process, I think it will be better to add the process of finding those specific values. I can also help to PR after your commitment.

apehex:
Hi @F01ivor4, thanks for the help!
I'm actually quite curious to go further too.
I'll try and do at least some of it and get back to you :)
BTW do you think the [bug report from april](https://github.com/one-hundred-proof/kyberswap-exploit) is relevant?

apehex:
Noob question: if i do another PR how can I include you in the process so that you can commit @F01ivor4 ?

F01ivor4:
Yes, it is relevant, I also read this before reproducing this attack. @apehex

PR #588, opened 2024-03-23
fix: error fix on actions

akshaynexus:


akshaynexus:
Build for all files dont work right now will cause this action to fail

remove it for now until codebase is fixed

it errors on a kyberswap exploit poc which was added recently

Add it as an issue to fix later on the codebase

PR #589, opened 2024-03-23
fix: fix final action issues

akshaynexus:


akshaynexus:
Ive tested the commands this runs locally and it works,should fix the action issue properly @SunWeb3Sec

SunWeb3Sec:
@akshaynexus fatal: ambiguous argument 'main': unknown revision or path not in the working tree.

PR #667, opened 2024-05-15
refactor: refactor exp files

akshaynexus:


SunWeb3Sec:
I think it's better not to modify someone else's POC？

akshaynexus:
It's mostly just refactor to use basetest for most of the poc,then some readability changes to the other few pocs

akshaynexus:
Let me know if any changes needed ,cause most changes are for readability and simplicity ,some changes such as logging balances on each chain,rearrangement of addrs based on the type of address like asset,contracts etc

think its ok but let me know

SunWeb3Sec:
This approach is good, but considering that beginners are more familiar to importing `forget-std/Test`

We can wait if there are no objections.

PR #784, opened 2024-07-22
Update 2024-06 info

Autosaida:


Autosaida:
Noticed that [this incident](https://github.com/SunWeb3Sec/DeFiHackLabs/blob/main/src/test/2024-06/XiaoPANG_exp.sol) was caused by someone mistakenly sending their LP token to the pair contract. This allowed anyone to burn the corresponding LP token and profit from it. So it just is a MEV activity, not a DeFi attack. Therefore, considering removing it, what are your thoughts? @SunWeb3Sec

SunWeb3Sec:
> Noticed that [this incident](https://github.com/SunWeb3Sec/DeFiHackLabs/blob/main/src/test/2024-06/XiaoPANG_exp.sol) was caused by someone mistakenly sending their LP token to the pair contract. This allowed anyone to burn the corresponding LP token and profit from it. So it just is a MEV activity, not a DeFi attack. Therefore, considering removing it, what are your thoughts? @SunWeb3Sec

yep, agreed.

Autosaida:
> > Noticed that [this incident](https://github.com/SunWeb3Sec/DeFiHackLabs/blob/main/src/test/2024-06/XiaoPANG_exp.sol) was caused by someone mistakenly sending their LP token to the pair contract. This allowed anyone to burn the corresponding LP token and profit from it. So it just is a MEV activity, not a DeFi attack. Therefore, considering removing it, what are your thoughts? @SunWeb3Sec
> 
> yep, agreed.

Removed it.

PR #792, opened 2024-08-06
Add POC for MAMO and EEE

shuo-young:
add mamo token exploit which caused about 3.3K USD


shuo-young:
Supplement the POC for EEE

shuo-young:
supplement the POC for EEE, which caused more than $22K loss

shuo-young:
change readme to the standard format according to the time

PR #814, opened 2024-09-20
Fix Zenterest link

robmcl4:
Zenterest is in the 2024-08 folder, not 2024-09


robmcl4:
reference: [src/test/2024-08/Zenterest_exp.sol](https://github.com/SunWeb3Sec/DeFiHackLabs/blob/main/src/test/2024-08/Zenterest_exp.sol)

PR #840, opened 2024-11-04
chore: add forge fmt to format solidity files

jsvisa:
use `forge fmt` to format and lint the solidity files.


SunWeb3Sec:
is it necessary?  need 2 more contributors approve it.

PR #880, opened 2025-03-27
Tmp update

QiLOL:
Not sure if we want to temp remove the affected link.
Refer:
- https://x.com/tayvano_/status/1904942559673397335?s=46&t=KiQtSeOezqgoqLCNIiK6LA
- https://x.com/tanuki42_/status/1905003045433290940


SunWeb3Sec:
I think yes!

Issue #883, opened 2025-03-31
Now ankr needs an API key to be used


rotcivegaf:
https://github.com/SunWeb3Sec/DeFiHackLabs/blob/926b25ffb528201cbbbf87c84ea60f1729dec1a8/foundry.toml#L8-L20

  Option 1) We could use an `.env` file where each one saves their API key.
  Option 2) We could update the list to public RPCs.

In both cases, I'm happy to help.

SunWeb3Sec:
any good public rpc for Option 2) ?

PR #885, opened 2025-04-01
Fix rpcs

rotcivegaf:
1. Update the list to public RPCs (#883 )
2. Add test to check RPCs health
3. Add linea and mantle RPCs
4. Change to from custom PRCs to RPCs defined in the list


SunWeb3Sec:
Looks good

PR #1184, opened 2026-07-16
Add Sodium ERC-4337 session-key validation-bypass PoC (~11.76 ETH, Arbitrum)

BhariGowda:
## Sodium  ERC-4337 session-key validation bypass (~11.76 ETH / ~$21.2k, Arbitrum)

**Root cause.** `_validateSignature` on the `executeWithSodiumAuthSession` path returns
valid whenever `sessionKey == signer` and no safe-session is set  it never verifies the
session-add auth proof (that check only runs later, in execution, not validation).

**Attack path.** Attacker signs userOps with its own EIP-1271 contract
(`isValidSignature` always returns true) and a self-chosen sessionKey, so
`validateUserOp` passes for ~300 wallets it doesn't own. Each op sets
`maxFeePerGas ≈ 4062 gwei` (~400,000× normal) with `callGasLimit=0` and deliberately
runs out of gas on execution, so EntryPoint charges ~0.5 ETH of each wallet's gas
deposit to the attacker (beneficiary) per op  ~11.76 ETH drained in one transaction.

**PoC.** The exploit tx is a CREATE (`to` is empty)  the entire attack runs inside the
constructor of a one-shot contract that loops all 300 victim wallets and self-destructs.
The PoC forks one block before the exploit, pranks the attacker (`tx.origin` AND
`msg.sender`, since the constructor guards on `ORIGIN`), and replays the exact creation
bytecode. Impact is asserted on the aggregate ETH gained by the attacker/beneficiary
(~11.76 ETH, matching the reported loss) rather than per-wallet deltas, since individual
wallets' deposits are not monotonic within the 300-op batch.

Exploit tx: `0x738995176c3bbd22f8deabd7a1e6b89a044231781b39aa2f350897535e6d7bc1`

Run: `forge test --contracts ./src/test/2026-07/Sodium_exp.sol -vvv --evm-version cancun`
(requires an archive Arbitrum RPC).


SunWeb3Sec:
🤖 **Automated PR review — MERGED**

✅ `forge test` PASS (`--evm-version cancun`, Arbitrum archive RPC)
- `testExploit()` passed — attacker ETH gained ≈ 11.76 ETH (11761122296529423844 wei), matching reported loss within tolerance
✅ README count: 833 → 834 (correct +1)
✅ TOC entry `[20260712 Sodium](#20260712-sodium---erc-4337-session-key-validation-bypass)` — anchor valid
✅ Date ordering: 20260712 correctly inserted between 20260714 and 20260701

PR #1194, opened 2026-07-28
add Projekt reward-vault self-dealing PoC (~301.7 ETH, Ethereum)

BhariGowda:
## Projekt (GREEN/GOLD) reward vault  permissionless purchase-tracking self-dealing (~301.7 ETH, Ethereum)

**Root cause.** The reward vault (`0x574fc478bc45ce144105fa44d98b4b2e4bd442cb`, unverified) credits
an ETH allocation via a permissionless `trackPurchase(buyer)` that sizes the reward from the buyer's
token balance delta  but never verifies real ETH was spent. Self-dealing balance changes therefore
inflate the payout arbitrarily.

**Attack path.** The attacker flash-loaned ~14K WETH from Morpho, pushed it into dozens of Uniswap V2
memecoin pairs (Kirby Inu, ROTTSCHILD, etc.) and `skim()`'d the tokens to its own contract, registering
huge fake "purchase" allocations for near-zero net cost (the flash loan is repaid in the same tx). It
then drained ~301.7 ETH from the vault's reward pool via `massWithdraw()` → `msg.sender.transfer(alloc)`.

**PoC.** The exploit tx is a CREATE (`to` empty)  the whole attack runs in the constructor. Two earlier
same-block txs seed the pair token excess the skims grab, so the PoC forks at the exploit tx (not
block-1) and deploys the exact creation bytecode, pranking the attacker as both `tx.origin` and
`msg.sender`. The vault is unverified, so the PoC asserts the reproduced drain: the vault's ETH reward
pool drops by ~301.7 ETH and that value is swept to the attacker's profit wallet.

Exploit tx: `0x90f40d3c3b60370f7287d51d972ef54596c46e98f21af91b03a4e84c5e410f64`
Attacker: `0x61e7ad696215688d274c729a4cd0fbbc88fc4f85`

Run: `forge test --contracts ./src/test/2026-07/ProjektRewardVault_exp.sol -vvv --evm-version cancun`
(requires an Ethereum archive RPC).


SunWeb3Sec:

- ✅ `forge test --contracts ./src/test/2026-07/ProjektRewardVault_exp.sol -vvv --evm-version cancun` — PASS
  - vault ETH drained: 301.704684684555317704
  - profit wallet gain: 300.704684684555317704
- ✅ README: incident count 840 → 841, new entry correctly hyperlinked and inserted newest-first above 20260724 Lien Finance, anchor resolves correctly.

Thanks for the contribution!

PR #1205, opened 2026-08-07
Add PoC: PantherBase Reality.eth governance timeout exploit, pre-production deployment no user funds (5.12M ZKP + 0.12 ETH, Aug 2026)

BhariGowda:
Reality.eth governance timeout exploit (5,124,773.63 ZKP + 0.1233 ETH, Base)

Note: this is Panther Protocol's Base deployment, which was not yet in production at the time of the attack. Per the team's own public statement, no live user funds were compromised. This is a governance-mechanism bug in a pre-launch deployment, included here as a reproducible attacker-controlled exploit pattern, not a live-funds incident.

Root cause, self-traced (no tx hash was publicly available  found by scanning Reality.eth's LogNewQuestion events on Base for the proposal, then following the Safe module's execution events): a Zodiac Reality.eth module connected to Panther's Base Safe was missing a safeguard (present on Panther's other deployments) that disables the module when there's no active governance proposal. Attacker submitted a Reality.eth proposal named "zkp-reexploit" to upgrade each ZKP proxy on Base to a malicious implementation, then immediately posted a "yes" answer with the required 0.5 ETH bond in the same transaction. No honest party posted a competing "no" bond within the 12h timeout + 8h cooldown window. The answer finalized unopposed, and the module executed 5 proxy upgrades, sweeping 5,124,773.63 ZKP + 0.1233 ETH to the attacker.

Confirmed genuinely permissionless: addProposal (propose) -> submitAnswer (bond, public) -> executeProposalWithIndex (execute once finalized, public). No signer, admin key, or code bug  the only precondition was that nobody contested the proposal in time. Same governance-capture class as the BarnBridge/CompoundProvider and StrongBlock PoCs already in this repo, but here the "capture" mechanism is a timeout-based optimistic oracle rather than a token-weighted vote.

PoC forks one block before the execution/drain transaction (the Reality.eth question is already finalized on-chain with result = yes at that point), replays the drain using the exact on-chain creation calldata as a hardcoded bytes constant. Asserts ZKP + ETH swept to the attacker's orchestrator contract, exact match to on-chain values.

Timeline: proposal + self-bond, then execution 76,338 seconds later (~21.2h = 12h timeout + 8h cooldown + slack), matching the reported mechanics exactly.

Attacker EOA: 0x7dB4cFea... (owns orchestrator 0x9400161d...)
Reality.eth module (Base): 0x4ce69e77... (clone of 0x4e35da39)
ZKP token (Base): 0x0a776C1c...
[...]

PR #1206, opened 2026-08-08
Add PoC: Atomic flash-loan price oracle manipulation of lending collateral valuation (~29,984 USDC, Aug 2026)

BhariGowda:
Flash-loan price oracle manipulation of lending collateral valuation (~29,984.27 USDC, Arbitrum)

Root cause, confirmed against the trace: a fresh EOA (nonce 1) called an attack contract that took an Aave V3 flash loan of ARB, then swapped against the Uniswap V3 ARB/USDC.e pool used by Atomic's strategy/AtomicLending module to value its concentrated-liquidity LP position and lending collateral. Under the skewed spot price, the strategy mis-valued the LP/collateral, letting the attacker unwind liquidity for more than they deposited. Extracted ARB was routed ARB->WETH->USDC and forwarded to the attacker EOA.

run() is gated only by a hardcoded msg.sender == attacker EOA check in the bytecode  no admin role, no signature, no owner-storage path. Every step (flash loan, swap, strategy unwind) is a plain permissionless call. Confirmed genuinely attacker-controlled, no privileged path anywhere in the trace.

Whole exploit is a single flash-loan-funded transaction  no pre-funding of the EOA, no same-block setup (the attack contract was deployed 596 blocks earlier). PoC forks at block-1, deploys the exact on-chain creation bytecode via raw CREATE, calls run() with the exact original calldata, pranks the attacker as both msg.sender and tx.origin. Asserts net attacker USDC profit  exact match to on-chain value (29,984.270865 USDC vs reported ~29,984).

Core vault/strategy contracts (0xf617a3ad.../0xc1b67703...) are unverified proxies; function/variable names are apparent root cause from calldata + trace, not confirmed source.

Attacker EOA: 0xf8803DaE13A6757E53711214769B5fb52Ec26C7E
Victim (unverified): 0x51ff48f2d43966be796692bdddfae96a435242a8
Tx: https://arbiscan.io/tx/0xbd4a009cd609a05f1a64458969a1e2c2065472f0ee06a322246f155be12e3a9a

Run: forge test --contracts src/test/2026-08/Atomic_exp.sol --evm-version cancun -vvv

Note: requires --evm-version cancun (Aave V3's flash loan implementation uses transient storage) and an archive-capable Arbitrum RPC.


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/Atomic_exp.sol --evm-version cancun -vvv` — PASS, exact profit match (29,984.270865 USDC)
- ✅ README: incident count updated to 852 (rebased onto main after the concurrently-merged PantherBase PR bumped it to 851 — the branch's original 850→851 diff was re-applied as 851→852), new anchor resolves correctly, entry correctly placed newest-date-first

Note: since this PR was opened before PR #1205 (PantherBase) merged, I rebased the branch onto updated `main` and pushed a follow-up commit fixing the incident count (851→852) to keep it accurate after both merges. No changes were made to the PoC itself.

Thanks for the contribution!

PR #1207, opened 2026-08-10
Add PoC: USM defund() price split-invariance rounding exploit (~70.83 ETH, Aug 2026)

BhariGowda:
defund() price split-invariance rounding exploit (~70.83 ETH, Ethereum)

Root cause, confirmed against the trace and reproduced against the real USM/FUM contracts: a pricing logic flaw in ethFromDefund() inside USM's defund() function. Single-redemption pricing uses the arithmetic mean of the current and estimated-final FUM sell prices, but this pricing lacks split invariance  combined with per-redemption state contraction (adjShrinkFactor) and integer rounding, calling defund() many times for small slices of the same total FUM amount returns more total ETH than one single large defund() call for the same total FUM.

Not a flash-loan or price-manipulation attack  this is a pure math/rounding exploit in how the redemption curve is evaluated across split vs unsplit calls.

Every defund() call in the attack sequence is a plain permissionless public call from the attacker's contract, invoked inside a Morpho flash loan callback. No signer, admin role, or privileged path anywhere in the trace. The flash loan is working capital only, not part of the exploit mechanism.

PoC forks one block before the exploit, replays the real attack sequence against the live USM/FUM contracts on mainnet (not a reimplementation of the pricing math)  fund() 11,579.98 ETH to mint FUM, then 64 defund() calls redeeming ~971,629 FUM each, driven by the attacker contract's live callback loop against actual minted balances rather than replayed frozen calldata. Assert net attacker ETH profit: 70.830977 ETH on-chain, exact match to the reported ~70.83 ETH.

Attacker EOA: 0xb92b2e47680c89da8f951b8963ef469f461a50fc
Attacker contract: 0x5a5e29ba89663a3558273354e990426f3cac7de7
Victim (USM): 0x2a7fff44c19f39468064ab5e5c304de01d591675
Tx: https://etherscan.io/tx/0xfae5e751b8ce01457cbb6b529839f24a0cff50faaabcbd0fd02ca0cf559b050e

Run: forge test --contracts src/test/2026-08/USM_exp.sol -vvv

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-852 incidents included.
+853 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -54,6 +54,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260809 USM](#20260809-usm---defund-price-split-invariance-rounding-exploit)
 [20260807 Atomic](#20260807-atomic---flash-loan-price-oracle-manipulation-of-lending-collateral-valuation)
 [20260806 UnistreetLaunchpad](#20260806-unistreetlaunchpad---arbitrary-call-injection-via-unvalidated-launch-forwarding)
 [20260806 PantherBase](#20260806-pantherbase---realityeth-governance-timeout-exploit-pre-production-base-deployment-no-user-funds)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/USM_exp.sol -vvv` — PASS
  ```
  Ran 1 test for src/test/2026-08/USM_exp.sol:USMExp
  [PASS] testExploit() (gas: 5509754)
  Logs:
    attacker ETH before: 0.000000000000000000
    attacker ETH after (net profit): 70.830977367697708372
  ```
- ✅ README: incident count incremented 852 → 853, new entry correctly hyperlinked with a resolving anchor (`#20260809-usm---defund-price-split-invariance-rounding-exploit`), and positioned correctly (newest-first, above the 20260807 Atomic entry).

Thanks for the contribution — merged!

PR #1213, opened 2026-08-22
Add PoC: SandboxOFT LayerZero delegate hijack via approveAndCall (10M SAND unbacked mint, Aug 2026)

BhariGowda:
LayerZero delegate hijack via approveAndCall (10,000,000 SAND minted unbacked, ~$517,170 face value, Base)

Root cause, confirmed against verified source and trace: OFTSand.approveAndCall(target, amount, data) is a generic arbitrary-call primitive. Its only guard is that the first 32-byte parameter of `data` must equal the caller (`doFirstParamEqualsAddress`)  this is meant to constrain who can spend an approval, but it never constrains WHICH function the token is coerced into calling. The token then executes `target.call(data)` as itself.

When `data` is set to `setDelegate(attackerAddress)`, the guard passes trivially  the first (and only) parameter of setDelegate IS the caller. The token then calls Endpoint.setDelegate(attacker) with msg.sender = the OFT token itself, hijacking the OApp's LayerZero delegate role (normally onlyOwner-gated) to the attacker.

From there, permissionlessly as delegate:
1. Endpoint.setConfig(OFTSand, ReceiveUln302, {requiredDVNs: [attacker]})  attacker becomes the sole required DVN (verifier) for the source chain.
2. As that DVN: ReceiveUln302.verify(header, payloadHash, 1)  attacker self-attests a forged inbound packet. The peer check passes because SAND OFT sits at the same contract address on Ethereum and Base, so peer[srcEid] == the OFT's own address == the packet's declared sender.
3. commitVerification finalizes the forged verification via Endpoint.verify.
4. Endpoint.lzReceive(...) triggers OFTSand._credit, a bare _mint(to, amount)  minting 10,000,000 SAND (1e25) to the attacker with zero backing.

Every step (approveAndCall, setDelegate, setConfig, verify, commitVerification, lzReceive) is a plain permissionless public call. No signer, admin key, or privileged path anywhere  the "delegate" and "DVN" roles are both attacker-hijacked through the arbitrary-call primitive, not through any credential compromise.

Note: this is one representative transaction from an ongoing campaign (400+ repeated mint transactions reported at time of alert, ~$49B cumulative face-value SAND minted). This PoC reproduces one clean, self-contained instance of the pattern rather than replaying the full campaign.

PoC forks Base at block-1, deploys the exact verbatim creation bytecode via raw CREATE (whole exploit runs in the attack contract's constructor), pranks the attacker as both msg.sender and tx.origin. Calldata (packet header, GUID, payloadHash, message, UlnConfig) is hardcoded from the real on-chain trace, not reconstructed. Asserts exactly 10,000,000 SAND (1e25) minted unbacked to the attacker, matching the on-chain result exactly.

Exploiter (this tx): 0x638Ccb18370eE228378a565c1d4D0F9620d7F296
Victim (SAND OFT): 0xac531Eb26Ca1d21b85126De8FB87E80E09002DcF
Tx: https://basescan.org/tx/0x76ed03844ff61520a0fb99278f92f2f1453b24ccbacd20b91131703e4a56a446

Run: forge test --contracts src/test/2026-08/SandboxOFT_exp.sol -vvv

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-856 incidents included.
+857 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -54,6 +54,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260822 SandboxOFT](#20260822-sandboxoft---layerzero-delegate-hijack-via-approveandcall)
 [20260820 FlashstakeV2](#20260820-flashstakev2---mispriced-reward-pool-instant-upfront-reward-extraction)
 [20260819 AllbridgeCCTP](#20260819-allbridgecctp---phantom-cctp-deposit-via-unverified-message-attestation)
 [20260815 FoxLpBondsPool](#20260815-foxlpbondspool---stale-_stakeamount-from-manipulable-amm-spot-quote)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/SandboxOFT_exp.sol -vvv` — PASS: `10,000,000` unbacked SAND minted to attacker, matching the reported on-chain result exactly.
- ✅ README: incident count incremented 856 → 857, new entry correctly placed newest-first with a resolving anchor.

Thanks for the contribution!

PR #1214, opened 2026-08-23
Add PoC: TermFinance TermMax meta-vault flashloan logic exploit (~$8.5M, Aug 2026)

BhariGowda:
TermMax meta-vault flashloan logic exploit (~2,841.74 WETH + ~1,679,639 USDC, ~$8.5M combined, Ethereum)

Note: initial public alerts described this as a DAO governance-token takeover (attacker buying cheap governance tokens and passing malicious proposals). That framing does not match the on-chain evidence: neither attacker wallet's transaction history shows any Governor contract interaction, vote token, or propose/vote/queue/execute sequence anywhere. The real mechanism is a TermMax meta-vault accounting/logic exploit driven by Aave V3 and Balancer V3 flash loans against Term Finance's vault system  no governance capture, no stolen key or admin bypass involved.

Two independent, coordinated drains:
1. ETH leg: attacker deployed a bespoke attack contract days ahead of time (funded via Tornado Cash for gas), seeded state against the tmvETH ETH Meta Vault and its sub-vaults (tsvShorewoodsETH, tsvAugustDigitalETH, tsvParityPrimeETH, tsvParityCoreETH), then triggered a single self-contained drain call funded by an Aave V3/Balancer V3 flash loan, extracting 2,841.74 WETH.
2. USDC leg: a second coordinated EOA ran an analogous exploit against the vault's stablecoin side, extracting ~1,679,639 USDC.

Both attack contracts still hold code at the fork block (not self-destructed), so both drains replay directly against real on-chain state. Both attacker EOAs' proceeds were consolidated to the same downstream address, converted to ETH and DAI  confirms both legs are part of the same coordinated exploit.

PoC forks one block before each drain (StrongBlock/BarnBridge-style  state already settled on-chain), replays the exact hardcoded entry calldata from each attacker EOA, asserts the balance delta for each leg. Both attack contracts are unverified; their entry points are treated as the on-chain-observed root cause from the trace, not confirmed source  noted in the file header.

Attacker A: 0xa908b3472d76e7744baB0A5911768a4a6300612B (attack contract 0x64E477800051EFb06Ae4086f4b258b270668b4dF)
Attacker B: 0x686457a7468B9B31c5dbA43b1b16077B48520691 (attack contract 0x4F4B614d...Eba17f)
Consolidation address: 0xD5183d8BfC65a50863C62aF2538198A8288FFc13

Run: forge test --contracts src/test/2026-08/TermFinance_exp.sol -vvv

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-857 incidents included.
+858 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -54,6 +54,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260823 TermFinance](#20260823-termfinance---termmax-meta-vault-flashloan-logic-exploit)
 [20260822 SandboxOFT](#20260822-sandboxoft---layerzero-delegate-hijack-via-approveandcall)
 [20260820 FlashstakeV2](#20260820-flashstakev2---mispriced-reward-pool-instant-upfront-reward-extraction)
 [20260819 AllbridgeCCTP](#20260819-allbridgecctp---phantom-cctp-deposit-via-unverified-message-attestation)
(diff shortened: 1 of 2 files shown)


BhariGowda:
Correction: my original root-cause framing in this PR description was wrong. I initially attributed this to a standalone flashloan logic bug in TermMax's vault accounting, based on the transaction trace not showing a Governor contract or a standard propose/vote/queue/execute flow.

The actual mechanism is governance capture (same family as StrongBlock/BarnBridge, already in this repo)  just via infrastructure I didn't initially check for. Term's vaults are governed through Aragon TokenVoting, with a voting token that wraps deposited vault shares (deposit alone gives no voting power). Almost nobody wrapped: the attacker held 90.66% of the wrapped voting supply on the ETH meta-vault and 100% on all five USDC sub-vault plugins. He opened a proposal disguised with the curator's routine "veto" title, self-voted, waited out the 6.04-day minimum duration, then executed through Aragon DAO -> Zodiac Delay/Roles modifiers -> the vault's Safe avatar to recall funds from real strategies and redirect the full balance to his own contract. The Aave V3/Balancer V3 flash loans are liquidity used to unwind the recalled strategy positions during execution, not the privilege-granting mechanism.

The mechanical PoC itself is unaffected  the fork/replay still works correctly because it forks right before execution, after the governance vote had already passed. I've pushed an updated header comment with the corrected root cause and on-chain-verified governance contract addresses (commit 13473fa2). Thanks for your patience while I got this right.

SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/TermFinance_exp.sol -vvv` — 2/2 passed
  - `testExploitETHVault`: 2841.743517563533112109 WETH drained (matches reported ~2,841.74 WETH)
  - `testExploitUSDCVault`: 1679639.260083 USDC drained (matches reported ~1,679,639 USDC)
- ✅ README: incident count incremented 857 → 858, new entry correctly hyperlinked with a resolving anchor, positioned newest-first.

Thanks for the contribution!

PR #1220, opened 2026-08-28
Add PoC: CashCowCoin privileged burn + sync() drains pair reserve via iterative sells (~165.47 WBNB, Aug 2026)

BhariGowda:
Privileged burn + sync() drains pair reserve via iterative sells (~165.47 WBNB, ~$117.4K, BSC)

Root cause, independently verified against the trace (not just the alert): CashCowCoin's trading proxy (0xf523...241c7, EIP-1967  read implementation slot back to confirm) delegatecalls into an unverified implementation (0x4287...29a9ac). Its sell() flow performs a normal PancakeSwap CCC->WBNB swap, then calls the CCC token's selector 0xb20a0b6f, which  confirmed directly from the decoded trace  transfers the full post-tax CCC amount straight out of the pair to the dead address, immediately followed by pair.sync(). This permanently locks in the pair's WBNB reserve at its already-reduced post-swap level while erasing the CCC that would have restored balance, so each sell cycle bleeds WBNB out of the pair with no recovery.

The attacker exploited this via 80 iterative sell cycles in a single transaction, monotonically draining the pair.

Confirmed permissionless: the whole exploit replays from a plain EOA prank with no admin/signer setup anywhere. The proxy's sell() entry point is public; the 0xb20a0b6f burn is gated to be callable by the proxy/router, but the proxy itself makes that call on behalf of any caller who invokes sell()  so the entire attack surface is reachable by anyone, not privileged access.

PoC forks one block before the exploit, replays the exact original calldata (selector 0x36b3d4df, 80 iterations) against the pre-deployed attack contract, pranks the attacker as both msg.sender and tx.origin  no logic reimplemented, the real proxy, CCC token, and pair all execute. Asserts the pair's WBNB reserve drain: on-chain state shows the pair going from 165.4899 to 0.0179 WBNB, with 165.4719 WBNB landing in the profit splitter  exact match to on-chain state at block 118384061, consistent with the reported ~$117.4K at ~$709/BNB.

Attacker EOA: 0x7977bdeee3a79dc85cc18739692e796b5d2513c4
Attack Contract: 0x7738b4d7c25e9a7092ae1ab402343b20340daeaf
Exploited Proxy: 0xf523224c6171f81c54b93f474ed4c78de91241c7
Vulnerable Implementation: 0x4287742e50fad6d3351000fd31632412ab29a9ac
Victim Pair (CCC/WBNB): 0x1dbe9458a6840784d5defd62c6b71386100097c0
Tx: https://bscscan.com/tx/0x89d8050641019a5a75fa3dafb4f64fb153e4dd30c0f1f51d06a6cc206d3ead43

Run: forge test --contracts src/test/2026-08/CashCowCoin_exp.sol --evm-version cancun -vvv

Note: requires --evm-version cancun (the attack contract's flash-loan/callback path uses transient storage) and an archive-capable BSC RPC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-862 incidents included.
+863 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -55,6 +55,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 
 ## List of Past DeFi Incidents
 [20260827 MoonwellMAMO](#20260827-moonwellmamo---chainlink-oracle-source-price-manipulation-of-mamo-collateral)
+[20260827 CashCowCoin](#20260827-cashcowcoin---privileged-burn--sync-drains-pair-reserve-via-iterative-sells)
 [20260825 EnjinCryptoItems](#20260825-enjincryptoitems---unprotected-registry-reinitialization---manager-takeover)
 [20260825 FHToken](#20260825-fhtoken---sell-tax-reserve-mismatch-via-premature-sync)
 [20260823 TermFinance](#20260823-termfinance---termmax-meta-vault-flashloan-logic-exploit)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/CashCowCoin_exp.sol --evm-version cancun -vvv` — 1 passed, 0 failed. Pair WBNB drained 165.489861... → 0.017933..., attacker splitter gained 165.471928... WBNB (matches PR claim of ~165.47 WBNB / ~$117.4K).
- ✅ README: incident count 862 → 863, correctly matches 1 new PoC. New TOC entry and detail header both inserted in correct newest-date-first position, anchor resolves correctly.
- No FFI required, no forbidden commands, no injection content found in PR body/code.

Thanks for the contribution!

PR #1221, opened 2026-08-29
Add PoC: AjnaFinance liquidation accounting manipulation via self-controlled auction (~124.8K USD cbETH pool, Aug 2026)

BhariGowda:
Liquidation accounting manipulation via self-controlled auction (~43.75 cbETH + ~1.5 WETH, ~$124.8K, Ethereum)

Ajna is a permissionless, oracle-free lending protocol by design  auction and bucket prices derive entirely from the pool's own internal state rather than an external price feed. This is a known architectural choice, not a bug in itself. The exploit abuses it: an attacker who controls both sides of a liquidation (their own over-leveraged borrower position AND the taker executing the liquidation) can shape the internal prices those sides interact at, then exit with more collateral value than the quote token they paid in.

Attack sequence for this instance (cbETH pool), traced from the sample tx and its preceding setup:
1. In earlier transactions, the attacker set up a borrower position, kicked it into liquidation, and seeded deposit into a high-price bucket (index 2000).
2. In the sample tx: bucketTake(borrower, false, 2000) mints the attacker bucket LP and auction collateral at the manipulated price, removeCollateral(max, 2000) pulls the cbETH out.
3. The whole sequence is wrapped in a Balancer flash loan (4x WETH) to fund a second bucketTake/take cycle, followed by AuctionSettle and repayDebt to close the attacker's own borrower position.
4. A Curve/UniV3 swap converts cbETH to WETH to repay the flash loan, retaining the surplus.

Every call (bucketTake, take, removeCollateral, repayDebt, flashLoan) is a plain permissionless public function on Ajna's pool contract. A `validate` staticcall observed in the trace is the attacker's own tx.origin authorization check on their own contract, not an Ajna-side gate. No admin, signer, or privileged path anywhere.

This is one of 7 pools hit by the same underlying bug class (per the reported breakdown: syrupUSDC $173.7K, wstETH $159.8K, rETH $127.4K+$15.6K, cbETH $124.8K+$12.1K, WBTC $101.8K, WETH/USDC $42.0K, sDAI $18.0K ~$775K total). This PoC faithfully reproduces the cbETH pool instance; replaying all 7 wasn't necessary since the mechanism is identical.

PoC forks one block before the exploit (real pool state, including the attacker's already-standing auction), replays the exact 196-byte original calldata against the pre-deployed attack contract, pranks the attacker as both msg.sender and tx.origin. Asserts net attacker gain: on-chain measured 43.7549 cbETH + WETH delta from 0.2150 to 1.7154, combined ≈ $124.8K, matching the reported cbETH-pool loss. A small delta vs the live on-chain figure (1.7154 vs 1.7194 WETH) comes from cross-block swap-pool drift when forking one block early  both round to the same reported figure.

Attackers: 0x6f2f5236b10fe7162da077a2779f8b5f04b7827e, 0xc213145ef56c0f162e0c3d79e6e107b25cb8c453, 0xcccc640018f8c2b00fa45f456017ad2378eb3447
Tx: https://etherscan.io/tx/0x12dfde527ef62882bfabb64362c9ae0e6bfb628363bd298d0d0956c9a114e4f5

Run: forge test --contracts src/test/2026-08/AjnaFinance_exp.sol -vvv

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-863 incidents included.
+864 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -54,6 +54,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260828 AjnaFinance](#20260828-ajnafinance---liquidation-accounting-manipulation-via-self-controlled-auction)
 [20260827 MoonwellMAMO](#20260827-moonwellmamo---chainlink-oracle-source-price-manipulation-of-mamo-collateral)
 [20260827 CashCowCoin](#20260827-cashcowcoin---privileged-burn--sync-drains-pair-reserve-via-iterative-sells)
 [20260825 EnjinCryptoItems](#20260825-enjincryptoitems---unprotected-registry-reinitialization---manager-takeover)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/AjnaFinance_exp.sol -vvv` — PASS (mainnet fork @ block 25,854,887)
  - cbETH drained: 43.754862224358830495
  - net gain (ETH-equiv): 45.255256247882230782
- ✅ README: incident count 863→864, new entry correctly anchored and placed newest-first in both the TOC and detail list.

Thanks for the contribution!

PR #1225, opened 2026-09-01
Add PoC: FloatProtocol Uniswap V3 spot-price manipulation of Hypervisor LP shares (~10.7066 ETH, Aug 2026)

BhariGowda:
Uniswap V3 spot-price manipulation of Hypervisor LP shares (~10.7066 ETH, ~$28K, Ethereum)

Root cause, confirmed against the trace: Float Protocol's Hypervisor contracts (Gamma/Visor-style concentrated liquidity vaults) price LP shares off getTotalAmounts(), which reads the underlying Uniswap V3 pool's instantaneous slot0 price with no TWAP check or slippage protection. The attacker flash-loaned 1,000 WETH from Morpho, swapped to skew the pool's tick, deposited into both Hypervisor contracts at the distorted valuation, swapped the tick back, then withdrew a richer token mix than deposited, repaid the flash loan, and kept the surplus.

Whole exploit is a single contract-creation transaction  the attack runs entirely inside the deployed contract's constructor. The only access check anywhere is tx.origin == attacker in the outer constructor (a self-guard, not a protocol-side restriction) plus Morpho's flash loan callback verifying its own caller. Hypervisor deposit()/withdraw() and the pool swaps are all plain public calls with the attacker as an arbitrary share receiver  no admin, signer, or owner path anywhere.

PoC forks at block-1, deploys the exact verbatim on-chain creation bytecode via raw CREATE (no logic reconstructed), pranks the attacker as both msg.sender and tx.origin. Asserts net attacker ETH gain  10.706591043820923462 ETH, exact match to on-chain state (12.2296 - 1.5230 ETH before/after), consistent with the reported ~10.71 ETH (~$28K). The attack's own bytecode hard-codes a 10.7 ETH minimum-profit floor before triggering the final selfdestruct payout.

Attacker: 0xaea29218262dc6b0904ca077f6527c49dfd426d9
Attack Contract: 0xb46655eb5b77de277063a75586d1883e951b6c54
Vulnerable Contracts (Hypervisor): 0x85cbed523459b7f6f81c11e710df969703a8a70c, 0xc86b1e7fa86834cac1468937cdd53ba3ccbc1153
Underlying Uniswap V3 Pool: 0xe8c2036068fc3b0161ee1def0e8d01df4eac0ac
Tx: https://etherscan.io/tx/0x3d7549db65344da2a41067e17791b17fac16ec6b8e5132e82e243f6541de5cff

Run: forge test --contracts src/test/2026-08/FloatProtocol_exp.sol --evm-version cancun -vvv

Note: requires --evm-version cancun (attack bytecode uses transient storage) and an archive-capable mainnet RPC.


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-08/FloatProtocol_exp.sol --evm-version cancun -vvv` — PASS
  ```
  attacker ETH before: 1.523042733552253695
  attacker ETH after: 12.229633777373177157
  attacker net gain (ETH): 10.706591043820923462
  ```
- ✅ README count incremented 866 → 867
- ✅ New entry hyperlinked with resolving anchor, correctly positioned newest-first

Thanks for the contribution!

PR #1230, opened 2026-09-05
fix: lowercase 99 broken past-incident TOC anchors and correct 3 display labels

BhariGowda:
README-only change. No PoC or source files are touched, so this exercises nothing new in CI.

The "List of Past DeFi Incidents" TOC links each pre-2026 entry to a heading in past/YYYY/README.md via a #fragment. 99 of those fragments carry mixed-case letters (e.g. #20241126-NFTG---access-control), but GitHub generates heading IDs in lowercase, so the fragment does not match the target heading and the link fails to jump. This lowercases those 99 fragments to match the actual generated IDs. Only the #fragment is changed; the visible link text (NFTG, X319, CoW, ...) is left as-is.

Also corrects 3 stale display labels where the link target was already correct but the shown text was not:
- 20260109: TOC label "TRU" -> "Truebit", matching the incident's detail heading and the repo's project-name convention
- Aztec Escape Hatch (2026-06): Contract link text "AztecEscapeHatch_exp.sol" -> "AztecEscapeHatch_exp2.sol", matching the href it already points to
- SKP (2026-05, second entry): Contract link text "SKP_exp.sol" -> "SKP_exp2.sol", matching the href it already points to

Verified after applying: all 126 in-page anchors resolve, external resolving anchors go from 612 to 711, the incident counter is unchanged, and no other anchor regressed.

A further 32 past-incident anchors have deeper text/structure mismatches (truncated or renamed headings, e.g. #20230328-safemoon-hack vs the heading id 20230328-safemoon-hack---access-control) that lowercasing alone does not fix. Those are intentionally left out of this PR and kept for a separate follow-up, since each needs individual heading lookup to derive the correct slug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 
 [20260110 FutureSwap](#20260110-futureswap---unit-mismatch)
 
-[20260109 TRU](#20260109-truebit---overflow)
+[20260109 Truebit](#20260109-truebit---overflow)
 
 [20260101 PRXVT](#20260101-prxvt---bussiness-logic-flaw)
 
@@ -633,7 +633,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 
 [20241203 Pledge](past/2024/README.md#20241203-pledge---access-control)
 
-[20241126 NFTG](past/2024/README.md#20241126-NFTG---access-control)
+[20241126 NFTG](past/2024/README.md#20241126-nftg---access-control)
 
 [20241124 Proxy_b7e1](past/2024/README.md#20241124-proxy_b7e1---logic-flaw)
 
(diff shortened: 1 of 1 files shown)


SunWeb3Sec:

Thanks for the cleanup! README-only change, validated and merged.

- ✅ Scope: only `README.md` touched (no PoC / source), so no `forge test` needed
- ✅ Incident count line unchanged (correct — no new PoC added)
- ✅ 99 past-incident TOC `#fragment` anchors lowercased to match GitHub's generated heading IDs — strictly safe (an uppercase fragment can never resolve); spot-checked `#20241126-nftg---access-control` and `#20241107-cow---access-control` against their real headings
- ✅ 3 display-label fixes verified against their existing targets: `TRU → Truebit` (heading `### 20260109 Truebit - OverFlow`), `AztecEscapeHatch_exp.sol → AztecEscapeHatch_exp2.sol`, `SKP_exp.sol → SKP_exp2.sol` (both hrefs already pointed at the `_exp2.sol` files)
- ✅ No file-path portions altered; changes confined to link text and `#fragment`

PR #1229, opened 2026-09-05
Add PoC: AnkrMORE ankrFLOW certificate over-mint via bond staking, drained through MORE Markets E-mode (~$410K, Aug 2026)

BhariGowda:
ankrFLOW certificate over-mint via bond-ratio staking, drained through MORE Markets E-mode (~15,488,124 WFLOW from the MORE reserve, ~$410K, Flow EVM)

Root cause, confirmed against the exploit trace and Ankr's verified source: Ankr's dual-token liquid staking on Flow EVM keeps two tokens for staked FLOW - ankrFLOWEVM (CertificateToken, value-accruing, ratio-priced; at the exploit block 1 cert ~= 1.19985 FLOW, ratio 0.833437e18) and aFLOWEVMb (BearingToken, rebasing, pegged 1:1 to FLOW, bond ratio 1e18). The staking pool (FlowStakingPool, extending Ankr's LiquidTokenStakingPool) mints the certificate that backs a freshly staked bond using the BOND share count instead of the certificate share count:

- _stakeBonds(staker, amount): shares = _bearingToken.bondsToShares(amount) -> with the bond ratio of 1e18 this is shares == amount
- _stake(...): _certificateToken.mint(address(_bearingToken), shares) mints that same count of the CERTIFICATE token

Because shares is bond-denominated (1:1 with the deposited FLOW), staking v FLOW as a bond mints v certificate to back it, but v certificate is worth ~1.19985*v FLOW - so the position is over-backed with ~20% excess certificate value out of thin air. The correct sibling path _stakeCerts scales by the certificate ratio (_certificateToken.bondsToShares(amount)); the bond path omits that scaling. BearingToken.unlockShares then burns the bond and hands the caller that over-minted certificate 1:1.

Amplifier (the E-mode lever): MORE Markets is an Aave-V3 fork. Both ankrFLOWEVM and WFLOW sit in E-mode category 1 ("Wrapped native tokens") at 97% LTV. The attacker supplies the over-minted certificate as E-mode collateral, borrows WFLOW near the certificate's inflated oracle value, unwraps and re-stakes, and loops - each round nets WFLOW because 0.97 * ~1.2 > 1 - until the WFLOW reserve is empty. Ankr's own FLOW reserve (~52,315 FLOW at the block) is never the bottleneck; the value is manufactured by the mint bug and cashed out of MORE's WFLOW reserve.

[...]

PR #1231, opened 2026-09-05
Add DHC exploit PoC - repeatable claim due to missing claimed-flag

BhariGowda:
DHC's award state machine has no re-claim guard. createAward() records a fixed reward with no per-award reserve or locked collateral. participateAward() doesn't require the award to still be unclaimed, so a claimed award (status=2) can be reset back to status=1 by anyone owning the position, for as little as 0-1 wei. claimAward() then pays the same fixed reward again from the shared proxy balance, with no elapsed-time check.

Confirmed empirically on-chain by reading the implementation and comparing owner vs attacker: the attacker is not the proxy owner, and the reset/claim path succeeded 18x on a single position with zero privileged access.

Attack was 4 attacker txs in block 120055460, not one self-contained call: contract creation, two setup calls (prime + initial pledge/participate), then a final call that runs the 18x claim loop and dumps the drained DHC for USDT on the PancakeSwap pair. PoC forks the parent block and replays all four txs in sequence, pranked as the attacker EOA, which reproduces the attacker's exact deployed contract address.

Realized profit is measured via the attacker's USDT delta (~71,851 USDT) rather than raw DHC amounts, since DHC is a reflection token and per-account DHC balances are tax-perturbed and not a clean loss figure. The proxy's DHC reserve drain (~568,371 DHC) is logged in the test as corroboration.

Attack tx: https://bscscan.com/tx/0xb4c89845d95fd721e6931dd223e24e8b78a9de8c3b4428251a36d9213e44c496


SunWeb3Sec:

Thanks for the contribution! Verified and merged.

- ✅ `forge test --contracts src/test/2026-09/DHC_exp.sol -vvv` → **PASS** (attacker USDT profit ~71,851, matching the report; the configured `bsc` RPC alias `binance.llamarpc.com` is dead/NXDOMAIN, so the reviewer temporarily repointed it to a public BSC archive endpoint to run the fork — a repo infra issue, not a PoC defect)
- ✅ README: count `870 → 871`, new `20260905 DHC` entry hyperlinked with a resolving anchor, correctly placed newest-first
- ℹ️ The PR was rebased/merged onto current `main` and the count re-applied as `870 → 871`: your original `869 → 870` collided with another PoC (AnkrMORE, #1229) that merged first and also bumped to 870, which would have left the counter one short.

```
[PASS] testExploit()
  Award proxy DHC reserve before: 568371.829157042339390458
  Attacker USDT profit: 71851.016820880579325777
```

PR #1232, opened 2026-09-06
Add Reddio exploit PoC - cross-share-class balance double-count via permissionless registerErc20

BhariGowda:
getTotalAssetBalance(asset) in the InvestmentManagerFacet sums provider-tracked legs plus getRemainingBalance(asset), which for a non-native asset is a raw IERC20(asset).balanceOf(address(this)) read with no per-share-class ledger. The Lido/ETH share class (rsvETH) reads this same raw stETH balance.

registerErc20(address) is permissionless  confirmed empirically via a live eth_call from a random non-owner address at the exploit's parent block, which succeeded (no revert) despite owner() resolving to a different address. This let the attacker register stETH themselves, creating a second share class (rsvstETH) whose deposits land in the exact same raw stETH balance rsvETH pricing reads from.

Confirmed flow, all in one CREATE tx (attack runs inside a self-destructing contract's constructor):

Flash-loan 1,139 WETH from Balancer
registerErc20(stETH)  permissionless
depositEth ~1,130 ETH  attacker acquires ~99% of rsvETH supply
Deposit ~9.34 stETH under the new stETH registration  inflates rsvETH's share price without minting any rsvETH, since it lands in the shared raw balance
Redeem rsvETH at the inflated rate  nets 1,139.5 ETH
Redeem rsvstETH  recovers the same stETH deposited in step 4
Swap stETH → ETH on Curve, repay the flash loan

Net attacker profit: 9.2525 ETH, matching the reported ~9.25 ETH.

PoC replays the exact on-chain creation bytecode via raw create(), pranked as the attacker EOA, forked at the parent block of the exploit tx.

Attack tx: https://etherscan.io/tx/0xe3cba90e865c6cba950ebce36a52607f51f1fd33cd9fb920c78803f19b57791a


SunWeb3Sec:
🤖 **Automated PR review — changes requested (flagged for human review)**

Thanks for the contribution! Holding this one for a maintainer instead of merging, because the PoC is a **verbatim replay of the on-chain exploit bytecode rather than a reconstruction of the exploit**.

- ❌ `Reddio_exp.sol` embeds the attacker's ~21 KB on-chain **creation bytecode** as a `hex"…"` blob, deploys it with `create(0, …)` under a pranked attacker EOA, and lets that opaque blob do the work. The file's own comments say it: *"replays the EXACT on-chain creation bytecode of the exploit tx (hardcoded verbatim, no attack logic reconstructed)."*
- ⚠️ Because that blob **is** the original transaction, `forge test` passing here only proves the tx replays — it doesn't demonstrate the vulnerability. A reader learns nothing about the root cause from an opaque bytecode blob, which defeats the purpose of this repo.

**What we need to merge this:** please reconstruct the exploit as readable Solidity — a named attacker/test contract that calls the victim's actual functions (the vulnerable selector, the swaps/deposits, the manipulation) so the mechanism is visible in the source. Keeping the incident write-up and addresses is great; it's the attack body that needs to be real Solidity, not a deployed bytecode blob (or a raw `TARGET.call(CALLDATA)` replay).

Once you push an update with a reconstructed PoC, the reviewer will re-run automatically.

SunWeb3Sec:

Thanks for the contribution! This PoC passed review and has been merged.

- ✅ **Authenticity:** genuine Solidity reconstruction — a named `ReddioExploit` contract drives the attack through the victim's real functions (`registerErc20`, `depositEth`, `depositErc20`, `manualWithdraw`) via a Balancer flash-loan callback, Lido `submit`, and a Curve swap. No bytecode/calldata replay.
- ✅ **`forge test`:** PASS (evm_version `cancun`, via the inline `forge-config` directive).
  ```
  [PASS] testExploit() (gas: 1817604)
    attacker net gain (ETH): 9.253090281137114614
  ```
- ✅ **README:** incident count `835 → 836`; TOC anchor resolves to the detail header; newest-date-first ordering correct.
- 🔧 **Auto-fixed:** inserted the missing blank line between the new TOC entry and its neighbor so it renders on its own line.

PR #1235, opened 2026-09-06
Add DHC exploit PoC - repeatable claim due to missing claimed-flag (reconstructed)

BhariGowda:
DHC's award state machine has no re-claim guard. createAward() (selector 0x7f200fee) records a fixed reward with no per-award reserve or locked collateral. participateAward() (0xe3db9b54) doesn't require the award to still be unclaimed, so a claimed award can be reset and re-participated for as little as 0-1 wei. claimAward() (0x43609f36) then pays the same fixed reward (1401.65 DHC) again from the shared proxy balance. Confirmed permissionless  attacker is not the proxy owner.

These four award-related selectors are unverified on the implementation contract; their names are behavior-inferred from calldata/trace analysis (create → returns award id → participate → claim, repeated) and called via explicit abi.encodeWithSelector with real arguments, not replayed as opaque calldata  noted honestly in the file's comments.

Real attack used a factory spawning parallel "puller" positions to parallelize the drain (412 total claims across positions, ~579,584 DHC drained) rather than one position looping  the factory/puller structure is a gas-optimization detail, not part of the vulnerability, so the PoC reconstructs the same economic mechanism via a single-position loop instead of mirroring the parallelization. Attacker seeded the position via a DODO flash loan of USDT swapped to DHC; the PoC simulates that seed with dealt USDT since the flash loan mechanism itself isn't the vulnerability, and uses the real PancakeRouter for the buy/sell legs since DHC is a reflection token.

Reproduced profit: 71,100.63 USDT (assertion allows a 2,000 USDT band around the reported ~71,851, since the real attack additionally dumped across multiple venues and paid an inviter cut on the initial pledge that this single-pair, single-position reconstruction doesn't fully mirror).

Attack tx: https://bscscan.com/tx/0xb4c89845d95fd721e6931dd223e24e8b78a9de8c3b4428251a36d9213e44c496

Resubmission of #1231 (reverted in the repo-wide bytecode-replay cleanup)  rebuilt as readable Solidity with real function calls, no bytecode blob or raw calldata replay.


SunWeb3Sec:

Thanks for the resubmission! This rebuilt PoC passed review and has been merged.

- ✅ **Authenticity:** genuine Solidity reconstruction, not a replay. A named `DHCExploit` contract drives the attack through visible selector calls built with `abi.encodeWithSelector(SELECTOR, realArgs)` (create → participate → claim), and the re-claim loop (`_participate(RESET_AMOUNT); _claim();`) expresses the missing-claimed-flag vulnerability directly. No opaque bytecode blob and no raw pre-encoded calldata replay.
- ✅ **`forge test`:** PASS. (The BSC RPC alias `binance.llamarpc.com` is dead — DNS NXDOMAIN — so the run was repointed to a BSC archive endpoint purely as infra; this is not a PoC defect.)
  ```
  [PASS] testExploit() (gas: 20236299)
    attacker net profit (USDT): 71100.634971777945699789
  ```
- ✅ **README:** rebased onto latest `main`; incident count `836 → 837`; TOC anchor resolves to the detail header; newest-date-first ordering correct alongside the same-day Reddio entry.
- 🔧 **Maintainer note:** resolved the README count/list conflict against `main` (the Reddio PoC merged first) and ensured blank-line separation between the new TOC entries.

PR #1239, opened 2026-09-08
Add SecuredFi exploit PoC - same-block self-trade manipulates getMarketUnitPrice to fabricate collateral

BhariGowda:
OrderBookLib.getMarketUnitPrice() returns blockTotalAmount * 10000 / blockTotalFutureValue whenever isReliableBlock is true  a flag that updateBlockUnitPriceHistory flips true as soon as same-block filled volume reaches minimumReliableAmount (or immediately on a fresh market with no prior history). There's no time-weighting. The collateral valuation path (FundManagementLogic._calculatePVFromFV) reads this price with isReadOnly=true, so it always takes the manipulated same-block ratio directly. Confirmed permissionless via trace  every call in the real exploit comes from a plain contract, no privileged role anywhere.

Real mechanism (confirmed against verified source, not just the incident writeup): a cross-account self-trade, not single-account — same-account lend+borrow nets to zero fillable position, so a single-contract replay of this attack silently drains nothing. The actual attack uses two real counterparties: a forwarder contract deposits WBTC and posts the LEND leg at par, while the main exploit contract posts the matching BORROW leg. With two genuine counterparties, the position books as a real fill, pinning getMarketUnitPrice near par and inflating a deliberately cheap earlier position into full collateral.

Reconstructed flow, all real typed calls: Balancer flash-loans 0.91 WBTC → deposit into the vault → phase-1 self-orders + withdraw → cross-account par self-trade (a named SecuredFiMaker contract posts LEND, the exploit contract posts the matching BORROW) → withdraw → phase-3 top-up orders → final withdraw(MAX) drains the pool → flash loan repaid. No bytecode blob, no raw calldata replay.

Reproduced: 0.89899004 WBTC net profit after flash repayment (pool drained 98.6%, from 0.91170963 to 0.01271959 WBTC)  matching the ~$72K single-drain figure. This is the largest single execution of the bug; the reported ~$104K total spans several separate, independent actors (an MEV frontrunner and copycat bots) exploiting the same publicly-observable vulnerability across multiple later transactions  those are downstream consequences of the bug becoming exploitable-by-anyone once revealed, not part of the vulnerability itself, so this PoC reconstructs the core mechanism only.

Attack tx (WBTC drain): https://etherscan.io/tx/0xcd6159860783d8d481cc4bf4dceddc88acdca6bb08f6445525b2904cfd75b1da
Victim pool: https://etherscan.io/address/0xb74749b2213916b1da3b869e41c7c57f1db69393

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 [20260905 Reddio - cross-share-class balance double-count via permissionless registerErc20](#20260905-reddio---cross-share-class-balance-double-count-via-permissionless-registererc20)
 
 [20260905 SecuredFi - same-block self-trade manipulates getMarketUnitPrice to fabricate collateral](#20260905-securedfi---same-block-self-trade-manipulates-getmarketunitprice-to-fabricate-collateral)
+
 [20260903 NotionalFinance - ERC1155 fCash mint overflow drains escrow](#20260903-notionalfinance---erc1155-fcash-mint-overflow-drains-escrow)
 
 [20260901 ReflexerGEB](#20260901-reflexergeb---shared-gebproxyactions-library-registered-as-safe-owner)
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-837 incidents included.
+838 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:
🤖 **Automated PR review — merged** ✅

Thanks for the contribution! All checks passed:

- ✅ **PoC authenticity**: genuine Solidity reconstruction — named `SecuredFiExploit` / `SecuredFiMaker` contracts making typed calls to the protocol's real functions (`executeOrder`/`deposit`/`withdraw`). No bytecode blob, no raw-calldata replay.
- ✅ **`forge test`**: PASS (default evm-version).

```
[PASS] testExploit() (gas: 10808092)
  TokenVault WBTC before: 0.91170963
  TokenVault WBTC after:  0.01271959
  attacker WBTC profit:   0.89899004
```

- ✅ **README**: incident count 840 → 841; new "Past DeFi Incidents" anchor resolves to the detail header; entry in correct newest-first position.
- 🔧 **Auto-fixed**: the new past-incidents entry was missing a blank line before the following entry (would have rendered mashed onto one line); a maintainer commit added the separator.

Merged. Thanks again!

PR #1240, opened 2026-09-08
Add RouterDrain exploit PoC - permissionless swap entry lets a fake V3 pool abuse router allowances

BhariGowda:
Confirmed on-chain via direct getter read: the router's factoryV3() returns 0x0, so its uniswapV3SwapCallback has no way to derive or verify a canonical pool address  any msg.sender is accepted. The swap entrypoint at selector 0x33411b5e is fully permissionless. The callback decodes a payer address from data the calling "pool" supplies, then calls transferFrom(payer, msg.sender, amount)  binding neither the caller to a real pool nor the payer to the actual swap initiator.

Attacker deploys a fake contract implementing only the V3 pool swap() selector (0x128acb08). It re-enters the router's own callback, naming any address that has ever granted the router an ERC20 approval as the payer  draining their standing approval regardless of whether they initiated anything.

The router additionally gates the entry call on balanceOf(msg.sender) >= amountIn without ever spending that balance, enabling a bootstrap-and-double pattern: seed a tiny amount, then repeatedly call with the attacker's current (growing) balance as amountIn, doubling the drainable amount each iteration until the victim is fully drained.

Reconstructed as a named RouterDrainExploit + FakePool pair making real typed calls  entry ABI recovered from trace and validated byte-for-byte against the real on-chain calldata (the router's own source is unverified, so this was necessary rather than assumed). No bytecode blob, no raw calldata replay. Requires evm_version = cancun (router uses transient storage).

Reproduced: 3 of the reported 29 victims fully drained to zero, matching real trace amounts (OCEAN, Kandura, SATURN  per this repo's convention that a representative subset is sufficient to demonstrate the pattern generalizes).

Attack tx: https://bscscan.com/tx/0x40eb22369da422a8275d5679054aa3a8c8906d93abc0bac3a3f2cad879389319
Vulnerable router: https://bscscan.com/address/0xa331fde028e6F17425AB9333c39ae43722340d24

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-841 incidents included.
+842 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260907 RouterDrain - permissionless swap entry lets a fake V3 pool abuse router allowances](#20260907-routerdrain---permissionless-swap-entry-lets-a-fake-v3-pool-abuse-router-allowances)
+
 [20260907 CozyFinance - unverified UMA Optimistic Oracle trigger enables unbacked protection-token redemption](#20260907-cozyfinance---unverified-uma-optimistic-oracle-trigger-enables-unbacked-protection-token-redemption)
 
 [20260905 DHC - repeatable claim due to missing claimed-flag](#20260905-dhc---repeatable-claim-due-to-missing-claimed-flag)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:
🤖 **Automated PR review — merged** ✅

Thanks for the contribution! All checks passed:

- ✅ **PoC authenticity**: genuine Solidity reconstruction — named `RouterDrainExploit` + `FakePool` contracts making typed calls to the router's real functions. The `.call` uses `abi.encodeWithSelector` with a typed `Route` struct under the recovered selector `0x33411b5e` (necessary because the router impl is unverified), not a raw-calldata blob or bytecode replay.
- ✅ **`forge test`**: PASS (evm_version cancun, set via the file's inline `forge-config` directive).

```
[PASS] testExploit() (gas: 3293748)
  victim token: OCEAN   -> 463712630900287455799
  victim token: Kandura -> 4136465745046839277137256
  victim token: SATURN  -> 29910904943654694204808
```

- ✅ **README**: new "Past DeFi Incidents" anchor resolves to the detail header; entry in correct newest-first position.
- 🔧 **Rebased onto latest `main`**: another PoC PR merged first, so I rebased this branch and re-applied the count bump (841 → **842**) — a plain merge would have silently left the count at 841 because both PRs edited the same count line. No changes to your PoC.

_Note: the CI archive-RPC alias for BSC was temporarily repointed locally to run the test (the repo's `bsc` alias host is currently unresolvable); `foundry.toml` was reverted and not modified in the merge._

Merged. Thanks again!

PR #1241, opened 2026-09-09
Add BalancerV1BPool exploit PoC - joinswapPoolAmountOut rounding drain across 5 pools (reconstructed)

BhariGowda:
BPool.joinswapPoolAmountOut lets the caller specify a desired BPT (pool token) output directly, while calcSingleInGivenPoolOut reverse-computes the required token input in 18-decimal fixed-point math. Confirmed against verified source: this path only enforces tokenAmountIn != 0, maxAmountIn, and MAX_IN_RATIO  no minimum-effective-input check. MIN_BALANCE only guards bind/rebind, never the join/swap math. Once a pool's WBTC reserve is compressed to ~2 satoshi via ordinary swaps, a large BPT mint's required WBTC input rounds down to 1 satoshi while still minting the full requested share.

Decoded from the real 290-call trace (richer than a simple compress-join-exit summary): 38 swapExactAmountIn + 3 swapExactAmountOut compress the pool's WBTC reserve to dust, then a loop of joinswapPoolAmountOut(WBTC, poolAmountOut, maxIn=1) (mint BPT for 1 satoshi) + exitswapPoolAmountIn(WBTC, poolAmountIn, minOut=1) (rebalance WBTC back to dust so the next join also rounds down) accumulates ~99% of the BPT supply, followed by 44 cleanup swaps and one final exitPool redeeming the accumulated share for a proportional cut of the pool's real DPI/USDC/WETH/WBTC reserves.

Reconstructed as a named BalancerV1BPoolAttack contract making real typed calls through all three phases. The real attack funded this with triple-nested flash loans (Spark/Aave, Morpho, Uniswap V3)  since the loan source isn't the vulnerability, working capital is provided via deal() instead, netting to zero once fully reclaimed in the final exit phase (disclosed plainly in the file). The join+exitSwap accumulation loop converges empirically in 322 iterations rather than the real tx's ~100 — the exact count isn't load-bearing for demonstrating the bug, only that the accumulation mechanism itself is real and typed.

Reproduced: attacker captures ≥99% of every original reserve (DPI, USDC, WETH, WBTC), pool driven to dust. Realized profit ~$110,975 (Chainlink-priced), matching the reported ~$110,839 for this pool instance. This pool is 1 of 5 hit by the same bug in separate transactions on the same day (aggregate ~$234K); this PoC covers only the DPI/USDC/WETH/WBTC pool named as the vulnerable contract.

Attack tx: https://etherscan.io/tx/0x72510b257cc09bde8435b83ac1636f9498ffc353583330600b5b8da43d0d1aff
Vulnerable pool: https://etherscan.io/address/0x2257aaac34bcb27900291f7b84ee2565a6cbac57

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-840 incidents included.
+841 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -65,6 +65,8 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 [20260901 ReflexerGEB](#20260901-reflexergeb---shared-gebproxyactions-library-registered-as-safe-owner)
 
 [20260831 FloatProtocol - Uniswap V3 spot price manipulation of Hypervisor LP shares](#20260831-floatprotocol---uniswap-v3-spot-price-manipulation-of-hypervisor-lp-shares)
+
+[20260831 BalancerV1BPool - joinswapPoolAmountOut rounding drain across 5 pools](#20260831-balancerv1bpool---joinswappoolamountout-rounding-drain-across-5-pools)
 
 [20260822 SandboxOFT](#20260822-sandboxoft---layerzero-delegate-hijack-via-approveandcall)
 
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ **PoC authenticity** — genuine Solidity reconstruction (named `BalancerV1BPoolAttack` making typed `swapExactAmountIn` / `joinswapPoolAmountOut` / `exitswapPoolAmountIn` / `exitPool` calls through all three phases); no bytecode blob or raw-calldata replay.
- ✅ **`forge test`** — passes:

```
[PASS] testExploit() (gas: 99502551)
  join iterations: 322
  profit USD: 110975.052542697949574504
```

- ✅ **README** — count 842 → 843; new entry hyperlinked with a resolving anchor, blank-line separated, and correctly ordered newest-date-first (20260831).

PR #1243, opened 2026-09-09
Add OrderFactory exploit PoC - missing access control on createOrderForBuyer drains buyer accounts

BhariGowda:
createOrderForBuyer (selector 0xbde886fc) is permissionless. Its only gate is getState(buyer) == 1 (active)  no msg.sender == buyer check, no buyer signature/nonce/authorization, and no validation of the attacker-supplied seller argument. Confirmed on-chain: in the real tx, the attacker's own helper contract makes every call, never the buyer.

The Account System moves the buyer's entire accountBalances(buyer) into a freshly created order proxy. Root cause confirmed by direct storage read (vm.load, not inferred): the order proxy's storage slot 27 gets set to the attacker-supplied seller address during initialization. abort(address) (selector 0x90cbfa19) authorizes draining against whatever address sits in that slot  so naming yourself as seller means your own abort call sweeps the full ETH balance out.

Reconstructed as a named OrderFactoryExploit contract (acting as the malicious seller) making real typed calls: createOrderForBuyer(buyer, address(this), ...) with the 9-arg tuple re-encoded and validated byte-for-byte against the real on-chain calldata (both the factory and account system are unverified, so this was necessary), the order proxy's address predicted via vm.computeCreateAddress exactly as an off-chain attacker would, then abort to drain. Real selectors (abort, getState, accountBalances) confirmed via keccak match. Requires evm_version = cancun (transient storage) and splits createForBuyer/doAbort into separate external calls (nesting them hit an InvalidJump).

Reproduced: all 5 buyers targeted in this tx fully drained, exact per-buyer amounts matching trace (10.05 / 5.05 / 3.05 / 1.547 / 0.55 ETH = 20.247 ETH total). The reported ~24.7 ETH is the full incident total; this PoC reproduces this one transaction.

Attack tx: https://etherscan.io/tx/0x201c7a9b4114c76fcda2b5de5d765505f4c5f7c194dddf505043501de6cbbb1a

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-842 incidents included.
+843 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ **PoC authenticity** — genuine Solidity reconstruction: named `OrderFactoryExploit` contract invokes the recovered `createOrderForBuyer` selector with real typed arguments (symbols/prices/params) and then the seller-gated `abort()`. No bytecode blob, no raw-calldata replay.
- ✅ **`forge test`** — passes (cancun via inline `forge-config`):

```
[PASS] testExploit() (gas: 10397770)
  drained from buyer: 10.05 / 5.05 / 3.05 / 1.547 / 0.55
  total ETH drained: 20.247000000000000000
```

- ✅ **README** — rebased/merged onto current `main`; count corrected to **844** (the intervening merge of #1241 had already taken 843), entry hyperlinked with resolving anchor and correctly placed newest-date-first (20260908, above 20260907).

PR #1245, opened 2026-09-09
Add AIC exploit PoC - pair skim / reserve-mismatch exploit, flash-swap leveraged (reconstructed)

BhariGowda:
AIC and NEX are both fee-on-transfer tokens, so a Pancake pair's real token balance can drift above its recorded reserves over time as fees accumulate as un-credited surplus. skim() is a permissionless PancakeSwap V2 function letting anyone extract the difference between a pair's real balance and its tracked reserve.

Both AIC and NEX are unverified on BscScan  root cause described from observed on-chain behavior, not confirmed source, and noted as such in the file.

Real sequence (recovered via a one-time forensic replay of the original attack's creation bytecode, then discarded  the shipped PoC contains only real typed function calls, no bytecode or calldata from that reference run): flash-swap the USDC/AIC pool's entire AIC balance minus 1, swap that AIC for NEX on the NEX/AIC pair, transfer most of the received NEX back into that same pair, call skim() (collapsing the pair's NEX reserve to effectively nothing while its AIC reserve stays near 83.26M) then sync(), use the small retained NEX sliver to buy the pair's entire skewed AIC reserve, repay the flash swap (borrowed amount × 10026/10000), dump the remaining ~40.17M AIC through USDC to WBNB, unwrap to the attacker EOA.

Reconstructed as a named AICAttack contract with a real pancakeCall flash-swap callback, real router swaps, real skim/sync calls, real repayment  no bytecode blob, no raw calldata replay, nothing running from replayed creation code.

Reproduced: 32.361267289208020008 BNB net attacker gain, matching the reported ~32.3609 BNB to within 0.0003 BNB (noise).

Attack tx: https://bscscan.com/tx/0x905cc861bcc525d3a8e699583943831b97500bbac11c92dc20ed6edbddd69f87

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-842 incidents included.
+843 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -71,6 +71,8 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 [20260831 FloatProtocol - Uniswap V3 spot price manipulation of Hypervisor LP shares](#20260831-floatprotocol---uniswap-v3-spot-price-manipulation-of-hypervisor-lp-shares)
 
 [20260822 SandboxOFT](#20260822-sandboxoft---layerzero-delegate-hijack-via-approveandcall)
+
+[20260803 AIC - pair skim / reserve-mismatch exploit, flash-swap leveraged](#20260803-aic---pair-skim--reserve-mismatch-exploit-flash-swap-leveraged)
 
 [20260809 USM](#20260809-usm---defund-price-split-invariance-rounding-exploit)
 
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ **PoC authenticity** — genuine Solidity reconstruction: named `AICAttack` contract runs flash-swap → router swaps → public `skim()`/`sync()` reserve-collapse → buy-out → repay as typed calls. No bytecode blob or raw-calldata replay.
- ✅ **`forge test`** — passes (BSC fork; the dead `bsc` RPC alias was temporarily repointed to an archival endpoint for verification and reverted afterward):

```
[PASS] testExploit() (gas: 1131388)
  attacker BNB gain: 32.361267289208020008
```

- ✅ **README** — merged onto current `main`; count corrected to **846**, entry hyperlinked with resolving anchor and reordered to correct newest-date-first slot (20260803, after USM 20260809 and before SummerFi 20260706).

PR #1246, opened 2026-09-09
Add EnsoFinance exploit PoC - short-window Uniswap V3 TWAP oracle enables ~9x overvalued strategy deposit

BhariGowda:
In Controller.deposit(), EnsoOracle's estimateStrategy() is called before and after the user's tokens transfer in, and shares are minted on the difference: mint = amountAdded * totalSupply / valueBefore. The valuation chain (EnsoOracle → ItemEstimator → ProtocolOracle.consult()) prices deposited tokens via Uniswap V3's pool.observe().

Confirmed by decoding the actual observe(uint32[]) calldata from the trace: the TWAP windows used are short and vary per pool (707-900s), producing near-spot rather than genuinely time-weighted pricing. Both the Controller and Oracle implementations are unverified on Etherscan, so I could not confirm the SlowMist alert's specific claim that a registry "fee" field is literally reused as the "secondsAgo" parameter  the observable, independently-reproduced fact is that the FARM token gets valued ~9x above its real Uniswap V2 price by this vault's V3-pool-based oracle, and that mispricing is what the PoC actually exploits (the forked oracle does the real mispricing itself, regardless of the exact root cause of the short window).

Real attack: swap 0.683 WETH for 268.42 FARM on Uniswap V2 (a separately, correctly-priced venue), then deposit that FARM into the strategy  the vault's own oracle credits it at ~6.3 WETH, minting inflated shares  then withdraw to realize the profit.

Reconstructed as a named EnsoExploit contract making real typed calls: wrap ETH, Uniswap V2 swap to FARM, deposit(strategy, router, 0, 0, abi.encode(Call[])) with a real settleTransferFrom call routed through Enso's GenericRouter, withdrawWETH with per-token transferFrom calls to pull the strategy's real UNI/AAVE/MKR holdings, then sell to WETH. deposit/withdrawWETH/settleTransferFrom selectors 4-byte-confirmed against the trace. No bytecode blob, no raw calldata replay.

Reproduced: 5.277 ETH realized profit. The reported ~5.6 ETH is the spot value of the looted UNI/AAVE/MKR tokens as held; this PoC sells them for a realized ETH figure, so the ~0.3 ETH gap is slippage from that liquidation, not a discrepancy in the exploit itself.

Attack tx: https://etherscan.io/tx/0x63fbfc4b47e810d604dbdab0db35b17366f421337d8c60955eb81cd5d6071ad3

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-842 incidents included.
+843 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -54,6 +54,8 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260909 EnsoFinance - short-window Uniswap V3 TWAP oracle enables ~9x overvalued strategy deposit](#20260909-ensofinance---short-window-uniswap-v3-twap-oracle-enables-9x-overvalued-strategy-deposit)
+
 [20260907 RouterDrain - permissionless swap entry lets a fake V3 pool abuse router allowances](#20260907-routerdrain---permissionless-swap-entry-lets-a-fake-v3-pool-abuse-router-allowances)
 
 [20260907 CozyFinance - unverified UMA Optimistic Oracle trigger enables unbacked protection-token redemption](#20260907-cozyfinance---unverified-uma-optimistic-oracle-trigger-enables-unbacked-protection-token-redemption)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ **PoC authenticity** — genuine Solidity reconstruction: named `EnsoExploit` contract buys FARM cheaply on Uniswap V2, deposits into the strategy where the short-window V3 TWAP oracle over-values it, then withdraws the strategy's real UNI/AAVE/MKR (typed `Call[]` structs via `abi.encodeWithSelector`) and sells them back. No bytecode blob or raw-calldata replay.
- ✅ **`forge test`** — passes (initial run hit a transient RPC 429; passed on retry):

```
[PASS] testExploit() (gas: 2474369)
  attacker profit (ETH): 5.277209113875295697
```

- ✅ **README** — merged onto current `main`; count corrected to **847**, entry hyperlinked with resolving anchor and placed newest-date-first at the top (20260909, above OrderFactory 20260908).

PR #1247, opened 2026-09-09
Add UnistreetLaunchpad exploit PoC - arbitrary call injection via unvalidated launch forwarding (reconstructed)

BhariGowda:
LaunchpadFactoryAuto.launch() is external payable with no role check. It forwards attacker-supplied initCalldata/modifyCalldata verbatim into Uniswap V4's PositionManager.multicall(), with the factory itself as msg.sender to the PositionManager. Confirmed directly from verified source, not assumed.

Critically, the factory is a shared custodian  it holds every launch's V4 LP position NFT, not just the caller's own. Since the injected calldata executes with the factory as msg.sender, an attacker can act on behalf of the factory against ANY custodied V4 position, including ones belonging to other users' launches.

Confirmed from the real trace: initCalldata = initializePool() on a legitimate-looking throwaway CPOC/WETH pool (cover for the real payload), modifyCalldata = setApprovalForAll(attacker, true)  granting the attacker operator rights over every position the factory custodies. The attacker then makes 7 separate modifyLiquidities calls directly (actions 0x0311 = BURN_POSITION + TAKE_PAIR) against 7 real victim token IDs (360162, 364347, 360385, 363137, 363145, 363678, 364293), sweeping both currencies out of each.

Reconstructed as a named UnistreetLaunchpadAttack contract making real typed calls  FACTORY.launch() with the real Params/SeedBuys tuples, the injected payload built via abi.encodeWithSelector against real PositionManager signatures (visible in the code, not opaque), FACTORY.predict(salt) for the deterministic pool key, then real POSM.modifyLiquidities burns against each real token ID. No bytecode blob, no raw calldata replay, no redeploying the original attack contract. Requires evm_version = cancun (Uniswap V4 transient storage).

Reproduced: 17,743.907229 USDC + 0.007209570881911319 WETH, matching reported values to the wei. The 7 drained positions also included several illiquid launch memecoin tokens, swept but left in the exploit contract and noted in a comment rather than forced into the assertion.

Attack tx: https://etherscan.io/tx/0x9583e95d5c88c7966e269197f4b09022f26b7a27ad2c13660dda6774e3136d14
Victim (LaunchpadFactoryAuto): https://etherscan.io/address/0xfb60cd0b36ad4bd839b91767a6ad9055ab6ad825

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-842 incidents included.
+843 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -71,6 +71,8 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 [20260831 FloatProtocol - Uniswap V3 spot price manipulation of Hypervisor LP shares](#20260831-floatprotocol---uniswap-v3-spot-price-manipulation-of-hypervisor-lp-shares)
 
 [20260822 SandboxOFT](#20260822-sandboxoft---layerzero-delegate-hijack-via-approveandcall)
+
+[20260806 UnistreetLaunchpad - arbitrary call injection via unvalidated launch forwarding](#20260806-unistreetlaunchpad---arbitrary-call-injection-via-unvalidated-launch-forwarding)
 
 [20260809 USM](#20260809-usm---defund-price-split-invariance-rounding-exploit)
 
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

- ✅ **PoC authenticity** — genuine Solidity reconstruction: named `UnistreetLaunchpadAttack` calls `FACTORY.launch()` with real typed `Params`/`SeedBuys` tuples, injects `setApprovalForAll(this,true)` as `modifyCalldata` (built visibly via `abi.encodeWithSelector`), then runs typed `POSM.modifyLiquidities` BURN_POSITION+TAKE_PAIR against the 7 custodied victim positions. No bytecode blob or raw-calldata replay.
- ✅ **`forge test`** — passes (cancun via inline `forge-config`):

```
[PASS] testExploit() (gas: 2589480)
  attacker USDC gain: 17743.907229
  attacker WETH gain: 0.007209570881911319
```

- ✅ **README** — merged onto current `main`; count corrected to **848**, entry hyperlinked with resolving anchor and reordered to correct newest-date-first slot (20260806, after USM 20260809 and before AIC 20260803).

PR #1248, opened 2026-09-10
Add AjnaFinance exploit PoC - liquidation accounting manipulation via self-controlled auction (reconstructed)

BhariGowda:
Ajna is a permissionless, oracle-free lending protocol by design  auction and bucket prices derive entirely from the pool's own internal state, not an external price feed. This is Ajna's known architectural choice, not a bug by itself. The exploit abuses it: an attacker who controls both sides of a liquidation (their own over-leveraged borrower position AND the taker executing the liquidation against it) can shape the internal prices those sides interact at, then exit with more collateral value than the quote token they paid in  since there's no external oracle to contradict attacker-positioned prices.

Confirmed from real source: bucketTake, take, removeCollateral, repayDebt, and flashLoan are all plain permissionless public functions. A validate staticcall visible in the trace is the attacker's own tx.origin authorization check on their own contract, not an Ajna-side gate.

Real sequence, recovered via trace against standing on-chain state (the borrower's auction and bucket 2000's ~49.34 WETH deposit were already live at fork, set up by the attacker's own earlier transactions): bucketTake(borrower, false, 2000) + removeCollateral(max, 2000) pulls an initial cbETH slice while dropping the borrower's outstanding debt; the attacker-controlled borrower then pulls the bulk of its freed collateral (~46.5 cbETH) via repayDebt(...), confirming "attacker controls both sides" directly on-chain  the borrower's own repay call, not the taker, delivers most of the drain.

Reconstructed as a named AjnaTaker contract (the liquidating side) making real typed calls, with the attacker-controlled borrower's leg driven via a prank of its standing position  no bytecode blob, no raw calldata replay of the original hardcoded payload.

Reproduced: 48.127 cbETH gained for 3.474 WETH spent (net ~44.65 ETH-equivalent, ~$124.8K), matching the reported cbETH-pool loss. Note: the real attack funded its take via a Balancer flash loan + cbETH→WETH swap; this PoC takes the economically-equivalent cash-funded path, so the exact token split differs slightly while net value matches  disclosed in the file.

This cbETH pool is 1 of 7 hit by the same bug class the same day (~$775K aggregate: syrupUSDC $173.7K, wstETH $159.8K, rETH $127.4K+$15.6K, cbETH $124.8K+$12.1K, WBTC $101.8K, WETH/USDC $42.0K, sDAI $18.0K). Replaying all 7 wasn't necessary since the mechanism is identical.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,11 +76,11 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 
 [20260831 BalancerV1BPool - joinswapPoolAmountOut rounding drain across 5 pools](#20260831-balancerv1bpool---joinswappoolamountout-rounding-drain-across-5-pools)
 
+[20260828 AjnaFinance - liquidation accounting manipulation via self-controlled auction](#20260828-ajnafinance---liquidation-accounting-manipulation-via-self-controlled-auction)
+
 [20260823 ArrakisGUNI - Uniswap V3 spot-price manipulation of vault mint/burn](#20260823-arrakisguni---uniswap-v3-spot-price-manipulation-of-vault-mintburn)
 
 [20260822 SandboxOFT](#20260822-sandboxoft---layerzero-delegate-hijack-via-approveandcall)
-
-[20260828 AjnaFinance - liquidation accounting manipulation via self-controlled auction](#20260828-ajnafinance---liquidation-accounting-manipulation-via-self-controlled-auction)
 
 [20260809 USM](#20260809-usm---defund-price-split-invariance-rounding-exploit)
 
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

Thanks for the contribution and the thorough write-up!

- ✅ **PoC authenticity:** genuine reconstruction — a named `AjnaTaker` contract making real typed Ajna calls (`bucketTake` → `removeCollateral`, then `take`), with the attacker-controlled borrower's `repayDebt` leg driven via a prank of its standing on-chain position. No bytecode blob, no raw-calldata replay.
- ✅ **`forge test`:** passes.
  ```
  [PASS] testExploit() (gas: 1030016)
    taker cbETH gained: 48.127374524782937205
    taker WETH spent:   3.474032191998137190
    net ETH-equivalent drained: 44.653342332784800015
  ```
  Net ~44.65 ETH-equivalent (~$124.8K), matching the reported cbETH-pool loss.
- ✅ **README:** incident count `848 → 849`; new entry hyperlinked with a resolving anchor.
- 🔧 **Auto-fixed:** the new entry was inserted below `20260822 SandboxOFT`, but `20260828` is newer — I moved it to its correct newest-first slot (between `20260831 BalancerV1BPool` and `20260823 ArrakisGUNI`) in both the "List of Past DeFi Incidents" TOC and the "List of DeFi Hacks & POCs" detail section, preserving blank-line separation. Committed and merged.

PR #1251, opened 2026-09-12
Add BeatXswap exploit PoC - slot0 spot-price oracle with no TWAP guard drains vesting BTX reserves

BhariGowda:
LiquidityVestingConvert._calculateQuote() reads the target Uniswap V3/PancakeSwap V3 pool's slot0().sqrtPriceX96 directly, squaring it for the spot price  no TWAP, no deviation guard. _executeMint() re-reads slot0() for the current tick. deposit() is confirmed permissionless (nonReentrant/whenNotPaused only).

The slippage guard is confirmed ineffective from source: amount1Min = requiredBTX * 97/100 validates the mint against the already-manipulated quote itself, not any independent price reference. An extra detail found during verification: only 25% of the deposited USDT reaches the LP (75% goes to treasury), yet the BTX side is sized off the FULL manipulated quote  so a small USDT deposit can drag out a disproportionately large BTX reserve. The minted LP NFT is recipient-set to the vesting contract itself.

Real attack, confirmed via trace: Moolah flash-loans 15,950 USDT; PancakeInfinity Vault flash-loans 6,000,000 BTX (via lock/lockAcquired/take). Attacker dumps BTX into the pool twice to crash sqrtPriceX96, calls deposit() on two separate vesting contracts (LiquidityVestingConvert and LiquidityVestingConvertOnce, identical logic, same target pool) while the price is manipulated  each mints an LP position using the vesting contract's own BTX reserves at the crashed tick. Attacker reverses the swaps, repays the BTX flash loan, sells the net-drained 2,984,557 BTX into an unrelated, unmanipulated PancakeInfinity CL pool for 77,512 USDT, repays the Moolah flash loan, nets 63,704.837356523361110351 USDT profit  reconciling both figures from the original alerts (they measure different points in the same attack, not conflicting numbers).

Reconstructed as a named BeatXswapExploit contract making real typed calls throughout  Moolah callback, Vault lock/take, pool swaps, real deposit() calls on both victims, reversal swap, CL-pool resale via CLPoolManager.swap. No bytecode blob, no raw calldata replay (only a hex"01" pool-echo swap-data byte and PoolKey struct fields appear as hex, neither is a blob). Requires evm_version = cancun (PancakeInfinity Vault uses transient storage) via an inline per-test directive, not a global config change.

Reproduced: 63,704.837356523361110351 USDT profit, matching the reported ~$63.7K to the wei.

Attack tx: https://bscscan.com/tx/0xcc71a3bb131c73462b0f25533070113a63c85e942a22e18bc945eec184eb5799


SunWeb3Sec:

- ✅ **Authenticity (Check 0):** genuine reconstruction. The oracle manipulation is performed on-fork — `POOL.swap` dumps BTX to crash `slot0().sqrtPriceX96`, then `deposit()` is called on both vesting victims at the manipulated price, all in a named `BeatXswapExploit` contract with real typed calls. No bytecode/calldata replay, no settled-state shortcut.
- ✅ **`forge test`:** PASS (evm_version cancun via the inline per-test directive).
  ```
  victim1 BTX reserve before: 1240403.093...  after: 0.0124...
  victim2 BTX reserve before: 1832090.213...  after: 0.0181...
  attacker USDT profit      : 63704.837356523361110351
  ```
- ✅ **README:** count 849 → 850; new "Past DeFi Incidents" entry hyperlinked with a resolving anchor; entry placed in newest-first position with blank-line separation, shortened to the project-name-only style now used in the list.

Rebased onto current `main` (which had moved after the MoonwellMAMO revert + the TOC-shortening commits) and merged. Thanks for the contribution!

PR #1252, opened 2026-09-12
Add UnifiProtocol exploit PoC - unregistered caller-supplied pool trusted for reward eligibility and math

BhariGowda:
The UP reward manager's mintUP(pool) calls pool.token() on a caller-supplied "pool" address and treats it as eligible whenever that returns WBNB  no factory check, no registry, no whitelist. claimUP then drives the reward calculation off pool.balanceOf(pool) and pool.totalSupply(), both read from the same caller-supplied contract.

The manager's source is unverified on BscScan; its interface was reconstructed from decoded trace selectors (mintUP, claimUP) rather than assumed. One empirical refinement found beyond the alert: mintUP reverts (Unifi: Invalid params) unless pool == msg.sender so the fake pool must present itself as the caller, but there's still zero validation of what that self-referencing contract actually is. This matches the real attacker, who deployed a single combined pool+exploit contract.

Reconstructed as two named contracts: UnifiFakePool implements token() returning WBNB, totalSupply() returning 1e18, and balanceOf() returning 6041339220182213708556 (the exact on-chain value, tuned to reach the manager's full UP balance without triggering an overshoot revert)  and UnifiProtocolExploit, which flash-swaps 0.001 WBNB, calls the real mintUP/claimUP entrypoints, redeems the drained UP via the real (verified) UPRedeemer.redeem, and repays the flash loan. No bytecode blob, no raw calldata replay.

Note from the original alert (context, unaffected by this PoC): treasury still holds ~9.17 BNB, the manager is not paused and still holds MINT_ROLE  this is a currently unpatched, re-exploitable bug.

Reproduced: 5.859108237446175701 BNB net profit, matching the on-chain value exactly.

Attack tx: https://bscscan.com/tx/0x74634e4ccc7e8922798c7043f839219b1a1a4d6087f57cdbf3bd1544198b02e8


SunWeb3Sec:

- ✅ **Authenticity (Check 0):** genuine reconstruction. The attacker's malicious contract is rebuilt as a named `UnifiFakePool` (fake `token0()==WBNB`, attacker-chosen `balanceOf`/`totalSupply`), deployed on-fork and used to drive the real `mintUP`/`claimUP` entrypoints via a real PancakeV2 flash swap, then redeemed at the real `UPRedeemer`. The hardcoded `balanceOf` is a chosen attack *parameter*, not a replayed blob — no bytecode/calldata replay, no settled-state shortcut.
- ✅ **`forge test`:** PASS (evm_version cancun via inline directive).
  ```
  manager UP balance before: 0.559116757634477509
  manager UP balance after : 0.000000000000000001
  attacker BNB profit      : 5.859108237446175701
  ```
- ✅ **README:** count 850 → 851; new entry hyperlinked with a resolving anchor, placed newest-first with blank-line separation, shortened to the project-name-only style.

PR #1253, opened 2026-09-12
Add OMNI404 exploit PoC - ERC-404 dual ID/amount transfer interpretation drains pool WETH

BhariGowda:
Two independent alerts described this incident with genuinely different root-cause framings. Verified both against OMNI404's actual deployed source — one alert had the real setup right but the wrong leak mechanism; the other correctly identified the real bug.

From verified source: transfer(to, valueOrId) treats any value ≤ maxTotalSupplyERC721 (50) as an NFT-ID transfer  but this path ALSO moves a fixed 1e18 fungible units alongside the NFT, regardless of the actual intended amount. When a Uniswap V3 pool executes an exact-output swap and calls transfer(recipient, amountOut) with a tiny wei-level amountOut (≤50), the token pays out a full 1e18 OMNI404 + an NFT while the pool's own accounting only debits that tiny wei amount.

The ID-path is normally guarded by require(msg.sender == _ownerOf[id]), and the whitelisted pool never legitimately owns any NFTs  so the attacker first opens a Uniswap V3 flash() borrowing 20 O404, which (via the protocol's normal floor-reconciliation minting on receipt, not a bug by itself) mints NFTs 1-21 to the attacker. Repaying that flash loan via 20 individual transfer(pool, id) calls (the ID-path) hands the pool ownership of those specific low IDs  at which point a loop of exact-output micro-swaps calling swap(true, -id) for id 1 through 21 each nets 1e18 O404 for ~2 wei WETH, since the pool's transfer call for each tiny payout hits the dual-interpretation ID-path.

One alert's ERC-404-whitelist setup and its reported NFT-count evidence (15→21 minted) both reproduced exactly; its stated cause (independent floor-rounding leaking across a buy/sell round-trip) was not what actually moved value. The other alert's dual ID/amount confusion is the real mechanism.

Reconstructed as a named OMNI404Exploit contract making real typed calls throughout: Balancer flash loan, an initial buy, three cycles of [V3 flash borrow → ID-path repay → micro-swap drain loop → partial sell], final sell to close. Real callbacks (receiveFlashLoan, uniswapV3FlashCallback, uniswapV3SwapCallback). No bytecode blob, no raw calldata replay.

Reproduced: 2.427 WETH from the run() leg alone  reconciles exactly with SlowMist's reported "2.4 WETH". The separately-reported ~3.02 ETH is run() plus three follow-up drain() transactions in the same 4-tx block (0.367 + 0.147 + 0.079 ETH); this PoC reproduces the core mechanism (the 3-cycle run() sequence) rather than all 4 transactions, per this repo's convention.

Attack tx: https://etherscan.io/tx/0x4cbc3d8db832eb5442ce1c11d79fda05cafabfe7906933ed25881d60bf41d6f3
Vulnerable pool: https://etherscan.io/address/0xb3f613b9bc84ddb29d78fa4685b01d98412bba0b


SunWeb3Sec:

- ✅ **Authenticity (Check 0):** genuine end-to-end reconstruction. The full multi-step attack runs on-fork in a named `OMNI404Exploit` contract: Balancer flash → per-cycle Uniswap V3 `flash()` that mints fresh NFTs → ID-path `transfer(pool, id)` repayment that leaves the pool owning the low NFT IDs → a loop of tiny exact-output micro-swaps that trigger the dual ID/amount interpretation (1e18 O404 paid for ~2 wei) → dump for WETH. The load-bearing setup (making the pool own low IDs) is performed, not assumed. No bytecode/calldata replay.
- ✅ **`forge test`:** PASS.
  ```
  pool WETH before: 3.295348945941616499   after: 0.868345565039939419
  O404 minted before: 15   after: 21
  attacker ETH profit: 2.427580402449730252
  ```
  (NFT mint 15→21 matches the reported evidence; profit reconciles with SlowMist's ~2.4 WETH for the run() leg.)
- ✅ **README:** count 851 → 852; new entry hyperlinked with a resolving anchor, newest-first with blank-line separation, project-name-only style.

PR #1254, opened 2026-09-12
Add EtherFiAtomicQueue exploit PoC - missing solver access control drains victim allowances

BhariGowda:
AtomicQueue.solve(offer, want, users[], runData, solver) has no access control on the caller-supplied solver parameter  no solver == msg.sender check, no signature, no registration or consent verification of any kind. Confirmed against verified source and real trace.

solve() executes want.safeTransferFrom(solver, user, assetsToUser), where assetsToUser is computed entirely from the attacker's own updateAtomicRequest() call (atomicPrice * offerAmount / 1e18). So naming any address that holds a standing want allowance to AtomicQueue as solver drains that allowance, paying the victim in whatever worthless token the attacker names as the "offer."

Real impact against real victims, not a self-contained demo. The attack tx drained liquidETH from 9 distinct real on-chain addresses, each via their own genuine pre-existing allowance to AtomicQueue. This PoC forks real mainnet state and drains those same real standing allowances  no mocked or synthetic victim. (SlowMist disclosed privately to ether.fi before going public; whether funds were later returned is separate from whether the drain itself moved real funds, which it did.)

Reconstructed as FakeOffer (a self-minted worthless offer token) + EtherFiAtomicQueueExploit, which for each real victim calls updateAtomicRequest() with atomicPrice computed dynamically from the victim's live on-chain balance/allowance (faithful to the real attack's logic, not hardcoded), then solve(offer, liquidETH, [self], "", victim). Cash-out uses the real Uniswap V4 mechanism across the same 3 pools the attacker used, correctly handling partial-fill settlement (take/settle against the pool's returned BalanceDelta, not the requested amountIn  a CurrencyNotSettled revert surfaced this and was fixed properly rather than worked around). No bytecode blob, no raw calldata replay.

Reproduced: 15.451368952027963380 ETH, within 0.015% of the on-chain 15.453645063405167626 ETH. The drained shares' NAV at the live accountant rate was ~15.96 ETH; the realized figure reflects real V4 swap slippage on cash-out, matching the on-chain attacker's actual result rather than the theoretical maximum.

Attack tx: https://etherscan.io/tx/0x7cbe0b4349513fed6d03ba8bf9ed708e10e07a501d10b5f344a25ae10595599b


SunWeb3Sec:

- ✅ **Authenticity (Check 0):** genuine reconstruction. The vulnerability — `AtomicQueue.solve()` has no access control on the caller-supplied `solver` — is exercised on-fork: the exploit mints a worthless `FakeOffer`, crafts a malicious `updateAtomicRequest` with an attacker-computed `atomicPrice` (derived dynamically from each victim's live balance/allowance), then calls `solve(offer, liquidETH, [self], "", victim)` to drain each victim's standing allowance, and cashes out via real Uniswap V4 swaps. The whole exploit action runs on-fork with real typed calls. The victims' pre-existing allowances are the legitimate vulnerable surface (not attacker-manufactured pre-fork state), and a normal user could not drain them — so this is a real attack, not a settled-state cash-out. No bytecode/calldata replay.
- ✅ **`forge test`:** PASS.
  ```
  liquidETH NAV rate: 1.104676255992308290
  attacker ETH profit: 15.451368952027963380
  ```
  (within 0.015% of the on-chain 15.4536 ETH; the gap is real V4 cash-out slippage.)
- ✅ **README:** count 852 → 853; new entry hyperlinked with a resolving anchor, newest-first with blank-line separation, project-name-only style.

PR #1257, opened 2026-09-13
Add GDC exploit PoC - chained EOA bypass, liquidity misclassification, burn mismatch, and reentrant tax-free exit

BhariGowda:
Four confirmed bugs chained together, verified against GDCToken's real source and a full local re-execution trace of the actual attack tx:

EOA-check bypass: isContract() uses extcodesize, checked in receive()/deposit()  but code size is 0 during a contract's own constructor, so helper contracts deposit while still deploying, satisfying an EOA-only guard.
Buy-as-liquidity-removal misclassification: a deposit sets addLiquidityUnlockTime[helper] > 0; any subsequent pair→helper transfer with that record set gets flagged isRemove = true, skipping BuyingProhibited()  letting the helper freely buy GDC through a direct pair.swap().
Burn-vs-actual-input mismatch: swapSellAward forwards only SELL_USER_PERCENT = 3000 (30%) of a sell to the pair as real input, but _burnSellAgainstPair(user, amount) burns the FULL nominal amount directly from the pair's own reserve and calls sync()  collapsing the reserve below what was genuinely deposited (221.24M GDC burned for free in the real attack).
Reentrant tax-free exit: the sell path sets swapping = true and skips the tax while that flag is set. _sendBNB (a raw .call) pays the seller's bound inviter mid-execution  a callback the attacker's own helper receives, letting it re-enter and dump a large pre-accumulated GDC position through a direct pair.swap(), untaxed, against the just-collapsed reserve.

Critically, the profit only exists across two coordinated windows, not from any single mechanism in isolation  confirmed empirically: window 1 alone (helper d1 injects 372 WBNB, pulls 230.83 back) is net-negative. Real profit comes from d1's large pre-bought bag (1.63e27 GDC for 365.56 WBNB) being dumped tax-free during the callback triggered by helper f95's sell, PLUS f95's own 30%-of-sell capture, PLUS the free WBNB backing of the 221.24M GDC burned by mechanism 3.

Full real choreography reconstructed: d1 deposits (constructor-phase, becomes the referrer) → f95 deposits and binds d1 as its inviter (effectiveReferralCount[d1] = 1) → f95 sells, firing the callback to d1 (which buys via direct swap, collapsing the reserve) → sweep f95's 30% real capture → f95 sells its last 1 GDC → the resulting callback triggers d1's full-bag dump. Funded via a real Moolah.flashLoan + onMoolahFlashLoan callback.

Every step is a real typed call  Moolah.flashLoan, bindReferrer, constructor-phase receive() deposits, direct pair.swap, live getAmountOut for exact-output sizing. No bytecode blob, no raw calldata replay, no settled-state shortcut.

Reproduced: 35.337151495976986490 BNB, exact to the wei against the real on-chain attacker's balance delta.

Attack tx: https://bscscan.com/tx/0xf12ccb683c51cc1c5907d362f3219b3a49a597be1fdfe1fb36bb2361bb3db877

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 **Reproduce DeFi hack incidents using Foundry.**
 
 
-853 incidents included.
+854 incidents included.
 
 Let's make Web3 secure! Join [Discord](https://discord.gg/Fjyngakf3h)
 
@@ -54,6 +54,8 @@ If you appreciate our work, please consider donating. Even a small amount helps 
 - [Giveth](https://giveth.io/donate/defihacklabs)
 
 ## List of Past DeFi Incidents
+[20260911 GDC](#20260911-gdc---chained-eoa-bypass-liquidity-misclassification-burn-mismatch-and-reentrant-tax-free-exit)
+
 [20260909 EtherFiAtomicQueue](#20260909-etherfiatomicqueue---missing-solver-access-control-drains-victim-allowances)
 
 [20260909 OMNI404](#20260909-omni404---erc-404-dual-idamount-transfer-interpretation-drains-pool-weth)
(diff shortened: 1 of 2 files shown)


SunWeb3Sec:

Everything checked out:

- ✅ **Authentic reconstruction** — the exploit is rebuilt as real typed calls (Moolah `flashLoan`, constructor-phase `receive()` deposits, `bindReferrer`, direct `pair.swap`, the referral `_sendBNB` reentrancy callback). No bytecode blob, no raw-calldata replay, and the manipulation (all four chained bugs) is performed on-fork rather than assumed from settled state.
- ✅ **`forge test` PASS** — `testExploit()` passed, `attacker BNB profit: 35.337151495976986490`, matching the on-chain attacker's delta to the wei.
- ✅ **README** — count incremented `853 → 854`; new `20260911 GDC` entry inserted newest-first in both lists with a valid, resolving anchor.

```
Ran 1 test for src/test/2026-09/GDC_exp.sol:GDC_exp
[PASS] testExploit() (gas: 4883423)
  attacker BNB profit: 35.337151495976986490
```

Thanks for the thorough write-up and the exact-to-the-wei reproduction!

PR #1258, opened 2026-09-13
Add ORB exploit PoC - leveraged pool buyout and refund-before-sell reentrancy drains burnLP-depleted reserve

BhariGowda:
ORBCore is whitelisted on ORBToken, so its sells skip the normal 5% burn tax. ORBToken.receive() calls into ORBCore.addPoolAndSell() with no reentrancy guard, and addPoolAndSell refunds BNB to the calling account BEFORE _sellToken completes  the attacker's own receive() immediately re-sends the refund, re-triggering the whole cycle (confirmed 58 levels deep in the real trace). Each iteration's _sellToken sells the account's ORB tax-free, then calls burnLP(sellBurnTax), which destroys ORB directly from the Pair's own reserve (capped at 20% of the pair's ORB balance per call) and calls sync().

Why the leverage phase is load-bearing, not decorative: the pair's real reserve is 45.3 WBNB / 5.826M ORB, and the attacker's actual pre-owned ORB inventory is just 3,483 tokens  seven orders of magnitude too small to meaningfully deplete a 5.8M-ORB reserve through the reentrant loop alone (a reentrancy-only reconstruction, matching a prior session's attempt, proves this: it's structurally unable to drain the pool). The real attack instead uses a large leveraged buyout to shrink ORB's side of the pool FIRST:

Outer Moolah flash-loan of 241,591 WBNB (its entire liquidity) + a nested Moolah flash-loan of 3,145 BTCB, deposited as Venus collateral to borrow 260,000 BNB  ~501k combined working capital.
A real router swapTokensForExactTokens(WBNB→ORB, to=Core) buys ~5.827M ORB out of the pool (routed through Core specifically  any non-Core buyer hits revert('not buy')), pushing the pool to ~441k WBNB / ~590 ORB.
The reentrant receive()→addPoolAndSell loop fires: descent swaps run against the still-full pool, the big buy executes at the deepest reentrancy level, and the unwind-phase burnLP calls then only need to grind down the now-tiny ~590 ORB remainder to dust  tractable where grinding the original 5.8M never was.
A final sell of the remaining ORB into the now near-empty-of-ORB pool extracts nearly the entire inflated WBNB reserve.
All leverage repaid in full (Venus repayBorrow+redeemUnderlying, both Moolah loans)  what remains is exactly the pool's original ~45 WBNB.

Reconstructed as a named exploit contract making real typed calls throughout: Moolah flashLoan/onMoolahFlashLoan, Venus enterMarkets/mint/borrow/repayBorrow/redeemUnderlying, real router and pair swaps, and the real receive()-driven reentrancy. No bytecode blob, no raw calldata replay, no settled-state shortcut. Requires evm_version = cancun (Venus's vToken uses transient storage), applied as an inline per-test directive.

Reproduced: 45.302 BNB net profit, matching the reported ~44.9-45 BNB, confirmed against the real on-chain pool reserve trajectory (45.318 → 0.34 → back to ~45 after repayment).

Attack tx: https://bscscan.com/tx/0x5e6b33b7d69b505d8ae6e50ca6967e13bf513b61593d29011dbcc6d134515b34


SunWeb3Sec:
🤖 **Automated PR review — changes requested**

Thanks for the contribution — the PoC itself is solid. It's an **authentic reconstruction** (real `Moolah.flashLoan`/`onMoolahFlashLoan`, nested BTCB loan, Venus `enterMarkets`/`mint`/`borrow`/`repayBorrow`/`redeemUnderlying`, real router `swapTokensForExactTokens` buyout, and the genuine `receive()`-driven reentrant descent with `burnLP` pool depletion). When run with cancun it passes cleanly:

```
[PASS] testExploit() (gas: 13419197)
  attacker BNB profit: 45.302221034270454423
```

Two things need fixing before this can be merged:

- ❌ **README `forge test` command is missing `--evm-version cancun`.** The command in the README is:
  ```sh
  forge test --contracts src/test/2026-09/ORB_exp.sol -vvv
  ```
  Run exactly as documented (repo default is `evm_version = 'shanghai'`), it **reverts** inside Venus's `getBorrowRate`/`mint` — the vToken uses transient storage (`TSTORE`/`TLOAD`), which isn't active pre-cancun. It only passes when `--evm-version cancun` is supplied. Every other transient-storage PoC in this repo carries the flag in its README command (e.g. `SummerFi_exp`, `DLMC_exp`, `JB_exp`, `WHALE_exp`). Please add `--evm-version cancun` to the README command for this entry so it reproduces as written. The PR description says cancun is "applied as an inline per-test directive," but there's no `forge-config`/`evm_version` directive in `ORB_exp.sol` — an inline `// forge-config: default.evm_version = "cancun"` above the test would also work if you'd rather keep the README command flag-free, but the flag-in-README approach matches the repo convention.

- ❌ **README now conflicts with `main`.** PR #1257 merged just ahead of this one and bumped the same count line + list, so this branch needs a rebase onto updated `main` with the count re-bumped (now `854 → 855`) and the `20260911 ORB`... entry re-applied in newest-date-first position.

- ✅ PoC is an authentic reconstruction and passes under cancun (~45.30 BNB, matching the reported loss).
- ✅ Anchor/entry formatting otherwise look correct.

Please push an update addressing the two items above and the reviewer will re-run automatically.

SunWeb3Sec:

Everything checked out:

- ✅ **PoC authenticity**: genuine reconstruction — a named `ORBExploit` contract making real typed calls (Moolah flash loans, Venus `enterMarkets`/`mint`/`borrow`/`repayBorrow`/`redeemUnderlying`, router/pair swaps, and the real `receive()`-driven reentrancy). No bytecode blob, no raw-calldata replay, and the leverage + manipulation are actually performed on-fork (not a settled-state shortcut).
- ✅ **`forge test`**: PASS with `--evm-version cancun`. `attacker BNB profit: 45.302221034270454423`, matching the reported ~45 BNB.
- ✅ **README**: count `854 → 855`, new entry hyperlinked with a resolving anchor, inserted newest-first with correct blank-line separation.

Thanks for the contribution and the thorough writeup!

PR #1259, opened 2026-09-14
Add YamFinance exploit PoC - dormant-DAO quorum governance takeover drains legacy UMA farming contracts

BhariGowda:
Yam Finance is a dormant DAO with very low active voting participation. Confirmed on-chain, YamGovernorAlpha's quorum is 2e29 and proposal threshold 5e28 (internal vote-weight scale, distinct from YAM's 18-decimal display). A proposal defeats only if forVotes <= againstVotes || forVotes < quorum no separate minimum "for" votes bar exists, so a large enough self-delegated holder clears both thresholds alone with zero other voters needed.

The attacker bought 504,427 YAM via an aggregator one block before delegating (real capital held across ~12 days of governance timing, not flash-loanable  confirmed via balance history: 0 YAM at block 25884983, 504,427 YAM at 25884986). Self-delegated weight of 2.016e29 alone cleared both quorum and threshold. Submitted YamGovernorAlpha proposal #45 (empty description) with a single action: setPendingAdmin(attacker) on the YAM Timelock. The vote passed and executed; the attacker called acceptAdmin(), becoming full Timelock admin over every YAM protocol contract and the DAO treasury.

From there: reduced the Timelock delay from 5 days to its 12-hour minimum, transferred gov of the legacy UMAFarmingMar/UMAFarmingFeb contracts to themselves via _setPendingGov/_acceptGov through the Timelock, called _settleExpired() on each (releasing WETH collateral from long-expired uGAS EMP positions  confirmed contractState = 2, settleable), then used the gov-only masterFallback function  a gated arbitrary target.call(data)  to drain both farms' WETH and UMA directly.

Reconstructed as a full real governance-timeline simulation, every step a real typed call: acquire and self-delegate real YAM, propose() (returns proposal ID 45, matching the real on-chain proposal exactly), castVote(true), advance the real 12,345-block voting period, queue(), advance the real 5-day timelock delay, execute(), acceptAdmin(), setDelay() to the minimum, the farming-contract gov transfer through the Timelock, _settleExpired() on both farms, and masterFallback calls draining each — the drain payload built via abi.encodeWithSignature("transfer(address,uint256)", attacker, balance), a genuine typed ERC20-transfer encoding demonstrating exactly how the escape hatch is abused, not a replay of captured calldata. No bytecode blob, no raw calldata replay, no settled-state shortcut.

Reproduced: 48.085 WETH + 763.101 UMA (23.499 WETH + 763.101 UMA from UMAFarmingMar, 24.586 WETH from UMAFarmingFeb), matching the reported ~$121K / 23.50+24.59 WETH + 763 UMA closely. The attacker's off-chain FixedFloat cash-out to 48.15 ETH is noted in the file header but out of the fork's scope.

Proposal-creation tx: https://etherscan.io/tx/0xf3c9b1d7094bd11e6aa065c5efb009afa682a961c72e3f4c1bc20e6a25d2e25a


SunWeb3Sec:

Everything checked out:

- ✅ **PoC authenticity**: genuine end-to-end reconstruction of the governance takeover — self-delegate → `propose` (returns proposal id **45**, matching on-chain) → `castVote` → advance the voting period → `queue` → warp the 5-day timelock → `execute` → `acceptAdmin` → reduce delay → transfer farm gov via the Timelock → `_acceptGov` → `_settleExpired` → `masterFallback` drain. Every step is a real typed call (no bytecode blob, no raw-calldata replay), and the drain payload is a genuine `abi.encodeWithSignature("transfer(address,uint256)", …)` encoding. Forking with the attacker's real pre-purchased YAM stake is the accepted "pre-existing balance" shape; the exploit itself is fully reconstructed on-fork, not read from settled state.
- ✅ **`forge test`**: PASS. `TOTAL WETH drained to attacker: 48.085375…`, `TOTAL UMA: 763.101265…`, matching the reported ~$121K.
- ✅ **README**: rebased onto latest `main` to resolve a count/list conflict with the just-merged ORB PoC; count `855 → 856`, new entry hyperlinked with a resolving anchor, inserted newest-first (20260912, above ORB) with correct blank-line separation.

Thanks for the contribution and the detailed writeup!

PR #1262, opened 2026-09-16
Add RsETHSafeModule exploit PoC - nested self-referential multicall authorization bypass drains Gnosis Safe via DELEGATECALL

BhariGowda:
multicall(address _contract, bytes[] _data) on the Router gates on _target, and _isAuthorized short-circuits true whenever _contract == address(this). Confirmed against real deployed bytecode and the full trace of the actual successful exploit tx.

The attack chains this trick twice: outer multicall(ROUTER, [...]) passes because the target is the Router itself, so the nested payload runs with the Router's own trusted identity  then an inner multicall(MODULE, [...]) is authorized because, from inside that trusted context, the enabled Safe module trusts the Router. The module then forwards execTransactionFromModuleReturnData(RecipeExecutor, 0, executeRecipe(...), operation=1)  a real DELEGATECALL into the victim Safe, executing a DeFi-Saver-style recipe that withdraws the Safe's Aave-deposited collateral.

Reconciled a genuine discrepancy across four independent alert sources, from real trace data rather than picking one to trust: the "original exploiter" addresses named in some alerts only appear in LATER blocks (25980849/25980871), attempting to drain an already-emptied Safe  failed copycat attempts, not the real successful attack. The actual successful exploit tx (block 25980525) is from MEV bot "yoink," which front-ran the original deployer's mempool-broadcast exploit and captured the value itself  same class of situation as this repo's SecuredFi PoC. Router and Safe module confirmed as two genuinely distinct contracts, not two different descriptions of one thing. Loss figures reconciled: 2,900 aEthrsETH is the gross amount irreversibly removed from the victim Safe; 2,882.37 rsETH (~$7.73M) is yoink's net realized amount after its own conversion-and-bribe leg.

Reconstructed as a named RsETHSafeModuleAttacker making real typed calls through both multicall layers via IMulticallRouter + abi.encodeCall, with the drain amount as a live typed parameter spliced into the recipe rather than a fixed replayed value. Includes a negative-control test confirming the identical module payload WITHOUT the self-referential wrap correctly reverts  isolating the actual bug in one assertion. No vm.etch/bytecode blob, no .call(rawCalldata) replay, no settled-state shortcut.

One disclosed fidelity limit: the Router, module, RecipeExecutor, and action executor are all unverified bespoke contracts (a private DeFi-Saver fork with mined selectors)  no verified source exists to type the exact recipe body against, so the recipe forwarded by the module is carried as three named data segments (head/mid/tail) with the typed drain amount spliced in. The authorization-bypass logic itself  the actual vulnerability  is fully typed throughout. The assertion targets the reconstructable, security-critical core: 2,900 aEthrsETH irreversibly removed from the Safe by an unauthorized caller, not the downstream MEV-race capture (out of scope, matching the SecuredFi precedent).

Example exploit tx: https://etherscan.io/tx/0x0e7680b06cb8a6f86c149d9ba90d98e3d334e7b072dde03909d43fcfd98a8705
Victim Safe: https://etherscan.io/address/0x40e93a52f6af9fcd3b476aedadd7feabd9f7aba8


SunWeb3Sec:
🤖 **Automated PR review — changes requested**

Thanks for the contribution! This PR can't be merged yet:

- ❌ `forge test --contracts src/test/2026-09/RsETHSafeModule_exp.sol -vvv` genuinely fails. The authorization-bypass logic itself does work as designed (the negative control `exploitWithoutSelfRef` reverts as expected, and the self-referential `exploit()` call does reach deep into the Safe — the module executes `execTransactionFromModuleReturnData`, the `RecipeExecutor` runs, and an ERC20 `approve` inside the Safe succeeds), but the recipe's later step — a call into the Uniswap V4 `PoolManager` (`modifyLiquidities`) — reverts with a custom error, and that failure is swallowed by the Safe module manager (`ExecutionFromModuleFailure`) rather than bubbling up. End result: `extracted ~2900 rsETH-equivalent: 0 !~= 2900000000000000000000`. Nothing is actually drained from the victim Safe in this reproduction.
- This does not look like an RPC/archive/evm-version infra issue — no `--evm-version` flag was needed or specified, and the early setup (fork, `isModuleEnabled`, balance reads) all succeed correctly against the real forked state.
- ✅ README updates look correct: incident count 857→858, new entry correctly anchored and ordered.
- ✅ Check 0 (PoC authenticity) passes — this is a real reconstruction of the auth-bypass logic via typed calls with a negative control, not a bytecode/calldata replay or a settled-state shortcut.

Relevant excerpt:
```
[FAIL: extracted ~2900 rsETH-equivalent: 0 !~= 2900000000000000000000 (max delta: 1000000000000000000, real delta: 2900000000000000000000)] testExploit() (gas: 888885659)
Logs:
  victim Safe aEthrsETH (before): 53402.655157601774360853
  aEthrsETH extracted from victim Safe: 0.000000000000000000
  victim Safe aEthrsETH (after): 53402.655157601774360853
...
    │   │   │   │   │   │   │   │   │   └─ ← [Revert] custom error 0x5bb354c4
    │   │   │   │   │   │   │   │   └─ ← [Revert] custom error 0x5bb354c4
    │   │   │   │   │   │   │   └─ ← [Revert] custom error 0xef3dcb2f
    │   │   │   │   │   ├─ emit ExecutionFromModuleFailure(param0: 0xeA18B13d11f705a68F0954f637949e1eaA7AC4ca)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped
```

Likely the spliced Uniswap V4 recipe payload (the `RC_HEAD`/`RC_MID`/`RC_TAIL` byte segments carrying the drain amount at two spliced offsets) doesn't reproduce the exact call the real `modifyLiquidities`/hook expects at this fork block — possibly a pool-state or hook-specific precondition not present in this reconstruction. Please debug the recipe forwarding step so the test actually extracts the collateral, then push an update — the reviewer will re-run automatically.

SunWeb3Sec:

This PR passes all checks and has been merged. Summary:

- ✅ **PoC authenticity (Check 0):** The self-referential `multicall` authorization bypass — the actual vulnerability — is fully reconstructed in typed Solidity (`IMulticallRouter` + `abi.encodeCall`), not replayed. The negative-control test (`exploitWithoutSelfRef`) proves the bypass is load-bearing by showing the identical payload reverts without the self-referential wrap. The only opaque data is the private DFS recipe body forwarded downstream, disclosed in the PoC's own comments — not the vulnerability logic itself.
- ✅ **`forge test`:** `PASS` — `testExploit()` (gas: 789745), ~2,900 aEthrsETH extracted from the victim Safe as expected.
- ⚠️ **README merge conflict:** Your branch's merge of `upstream/main` left literal unresolved conflict markers (`<<<<<<<`/`=======`/`>>>>>>>`) in `README.md` around the incident-list entries (this happened because another PR — BonfireSwap — merged to `main` first, based on the same date). I resolved this directly on your branch, keeping both entries and bumping the incident count to 859. No changes were made to your PoC source.

Thanks for the detailed writeup and the negative-control test — that's exactly the kind of evidence that makes an auth-bypass PoC easy to verify.

PR #1264, opened 2026-09-16
Add BonfireSwap exploit PoC - missing caller authorization lets anyone force-sell any approved holders tokens

BhariGowda:
BonfireSwap.transfer(address to, uint amountAIn, address beneficiary, uint deadline) calls _safeTransferFrom(tokenAddress, to, pancakePair, amountAIn)  a transferFrom(from=to, ...)  with no check that msg.sender == to and no verification the caller holds any allowance of their own. It spends purely on to's own pre-existing allowance to the router. simpleTransfer/loggedTransfer share the identical flaw. Confirmed against verified source (Bonfire.sol:100) and live on-fork state  every real victim had unlimited (MAX) allowance to the router.

One correction to both original alert sources: skimPools inside transfer() actually refunds the freshly-force-sold BONFIRE back to beneficiary, not WBNB  each transfer() call deposits the named victim's tokens directly into the attacker as BONFIRE, not cash. The standalone skimPool(pair, ...) call the attacker also makes is wrapped in try/catch and reverts (INSUFFICIENT_OUTPUT)  it doesn't realize any value, contrary to what both alerts implied. The attacker actually monetized by calling the router's own sell() on the accumulated BONFIRE afterward (confirmed via real WBNB flow in the trace: two sells, 33.22 + 32.86 WBNB).

The 65-vs-41 victim count discrepancy between sources is explained by 67 raw addresses in the transfer logs including the attack contract itself and the token contract's own reflection/tax bookkeeping entries  ~65 real external holders, ~41 with balances worth draining. The largest real victim and drained amount match the SlowMist alert exactly.

Reconstructed as a named BonfireForcedSeller contract making real typed calls (transfer, sell, approve) against the real router, token, and pair on a live fork  using the 5 largest real victims with their real on-chain allowances (queried live, not invented) and the exact real drain amounts, including replicating the token's real _maxTxAmount = 5000e18 cap, which forced the real attacker to split the top victim's drain into two chunks  the PoC does the same. No bytecode blob, no raw calldata replay, no settled-state shortcut.

These 5 victims represent 5,017 of the 5,082 BONFIRE the attacker actually accumulated (~98.7%). Reproduced: 58.78 BNB (~89% of the reported ~66 BNB total)  the remaining gap attributed to chunking/slippage/reflection-tax effects in the sell leg, stated plainly rather than forced to match exactly.

Attack tx: https://bscscan.com/tx/0xb4c00e8f3ba815b6c70f45026f8794d2c1f079646a89919077688ce60692193f


SunWeb3Sec:

- ✅ `forge test --contracts src/test/2026-09/BonfireSwap_exp.sol -vvv` passes (after temporarily repointing the dead `bsc` RPC alias to an archive endpoint locally for testing only — `foundry.toml` in this repo is unchanged): `[PASS] testExploit() — BNB extracted (5-victim subset ~98.7% of drain): 58.780939117409582177`.
- ✅ PoC is a genuine reconstruction of the missing-authorization bug (attacker calls the vulnerable `transfer()` directly against victims' real pre-existing router allowances) — not a bytecode/calldata replay.
- ✅ README: incident count 857→858, new entry correctly anchored and ordered.

Thanks for the contribution!

PR #1250, opened 2026-09-11
Add ZentraFinance exploit PoC - aToken burn safeguard clamps to zero, repayWithATokens clears debt without burning

BhariGowda:
Zentra's lending core is a fork of Aave V3 Core v3.0.x. This is NOT the publicly-disclosed March 2026 upstream Aave V3 rounding vulnerability  it's a separate boundary condition in Zentra's own deployed aToken implementation, confirmed directly against the deployed source down to the exact lines.

_burnScaled scales via rayDivCeil, then clamps: if (amountScaled > scaledBalance) { amountScaled = scaledBalance; }  with no require(amountScaled - scaledBalance <= 1). The deployed source's own comment admits this clamp was meant to absorb only a ≤1-wei ceil overshoot on withdraw/liquidation. If the caller holds zero aTokens, the clamp reduces the burn all the way to zero  _burn(user, 0) no-ops, doesn't revert, while Pool.repayWithATokens records the full debt as repaid.

Confirmed via real trace: log 20 clears the full variableDebt (140000000001); log 25 "burns" zctUSD from an account that never held any (the emitted Burn event still carries the requested amount even though nothing was destroyed); log 26 records the Repay as complete with useATokens=true. Debt cleared, nothing burned, full collateral withdrawable.

repayWithATokens confirmed permissionless in the Pool implementation  public virtual override, no role or modifier, onBehalfOf = msg.sender.

Real attack: flash 200,000 USDC.e from an Algebra DEX pool (ctUSD/USDC.e, 0.01% tier  not Uniswap as initially assumed), post as collateral, borrow ctUSD, call repayWithATokens for debt + exactly 1 base unit (chosen to trigger ceil-rounding) while holding zero aTokens, withdraw the original collateral, repay the flash loan.

Reconstructed as a named exploit contract making real typed calls  no bytecode blob, no raw calldata replay. Requires evm_version = cancun (Citrea uses transient storage), applied via an inline per-test directive rather than a global config change.

Reproduced: 139,999.999999 ctUSD drained + 9.999999 USDC.e net of flash fee, matching the on-chain attacker's receipt to the base unit.

Note: the tx hash provided in the original alert (0x31ca0d...a8ea94) returns null on both the Citrea mainnet node and Blockscout  the real exploit tx is 0x9ac5df7e...074aa1 (block 12428145), documented in the file header.

Added the citrea RPC alias to foundry.toml (genuinely missing chain support, not a repoint of an existing alias).

[...]

PR #1265, opened 2026-09-17
Add ZentraFinance exploit PoC - aToken burn safeguard clamps to zero, repayWithATokens clears debt without burning

BhariGowda:
Zentra's lending core is a fork of Aave V3 Core v3.0.x. This is a boundary condition in Zentra's own deployed aToken implementation, confirmed against the deployed source down to the exact lines. Not the publicly-disclosed March 2026 upstream Aave V3 rounding issue.

_burnScaled scales via rayDivCeil, then clamps: if (amountScaled > scaledBalance) { amountScaled = scaledBalance; } with no require(amountScaled - scaledBalance <= 1). The deployed source's own comment says this clamp was meant to absorb only a ≤1-wei ceil overshoot on withdraw/liquidation. If the caller holds zero aTokens, the clamp reduces the burn to zero and _burn(user, 0) no-ops without reverting, while Pool.repayWithATokens still records the full debt as repaid.

Confirmed via real trace: log 20 clears the full variableDebt (140000000001), log 25 burns zctUSD from an account that never held any (the Burn event still carries the requested amount even though nothing was destroyed), log 26 records the Repay as complete with useATokens=true. Debt cleared, nothing burned, full collateral withdrawable.

repayWithATokens is permissionless in the Pool implementation, no role or modifier, onBehalfOf = msg.sender.

Real attack: flash 200,000 USDC.e from an Algebra DEX pool (ctUSD/USDC.e, 0.01% tier), post as collateral, borrow ctUSD, call repayWithATokens for debt + 1 base unit while holding zero aTokens, withdraw the original collateral, repay the flash loan.

Reconstructed as a named exploit contract making real typed calls, no bytecode blob and no raw calldata replay. Requires evm_version = cancun since Citrea uses transient storage, applied via an inline per-test directive rather than a global config change.

Reproduced 139,999.999999 ctUSD drained plus 9.999999 USDC.e net of flash fee, matching the on-chain attacker's receipt to the base unit.

Note the tx hash in the original alert (0x31ca0d...a8ea94) returns null on both the Citrea mainnet node and Blockscout. The real exploit tx is 0x9ac5df7e...074aa1, block 12428145, documented in the file header.

This is a re-push of #1250 scoped to just the PoC and README. That PR also added a citrea RPC alias to foundry.toml, which put it outside the auto-review file scope. Left the RPC alias out here since the bot can't merge foundry.toml changes, so the PoC can be run with the citrea fork endpoint set via FOUNDRY_RPC_ENDPOINTS locally rather than committed.

Post-mortem: https://x.com/zentrafinance/status/2098415552293195831


SunWeb3Sec:

This PR passes all checks and has been merged. Summary:

- ✅ **PoC authenticity (Check 0):** The full causal chain is reconstructed and executed on-fork — flash loan, supply collateral, borrow, then `repayWithATokens` while genuinely holding zero aTokens (triggering the vulnerable zero-clamp in `_burnScaled`), withdraw, repeat for the second leg, repay the flash. No bytecode blob, no raw-calldata replay, no settled-state shortcut — the exploit precondition (holding zero aTokens) is created and the vulnerable function is actually called against it.
- ✅ **`forge test`:** `PASS` — `testExploit()` (gas: 2,004,267); ctUSD drained 139,999.999999, USDC.e kept 9.999999 — matching the PR's reported on-chain figures to the base unit.
- ℹ️ **RPC alias:** Added a temporary `citrea` alias to `foundry.toml` to run the fork (a public endpoint I selected, not one from the PR), then reverted it after verification — per your note, this stays out of the committed diff.
- ⚠️ **README count conflict:** Your branch's baseline (858) was superseded by PR #1262 merging first (858→859), so a straight merge would have silently left the count at 859 instead of 860. I rebased your branch onto the latest `main` and corrected the count to 860 before merging. No changes were made to your PoC source.

Thanks for the clear root-cause writeup and for re-scoping this out of `foundry.toml` to keep it in auto-reviewable scope.

PR #1266, opened 2026-09-18
Add Flamincome VaultYUSDT exploit PoC - permissionless Convex stakeFor inflates Strategy NAV, VaultYUSDT overpays on withdrawAll

BhariGowda:
VaultYUSDT.balance() -> Strategy.balanceOfY() -> impl 0xFf20DE3f.deposited() values the Strategy's Convex position as BaseRewardPool.balanceOf(Strategy) x metapool.get_virtual_price() plus aUSDT and USDT reserves. Two compounding flaws combine to drain it.

First, Convex's BaseRewardPool.stakeFor(Strategy, amount) is permissionless. Any address can stake Convex deposit tokens "for" the Strategy, directly inflating what the Strategy believes it owns without the Strategy ever depositing or consenting to anything.

Second, the Convex pool in question is the USDP/3CRV metapool, which was deeply depegged at the time (1,526,591 USDP vs 3,263 3CRV, USDP trading around $0.20), yet get_virtual_price() still reported around 1.0149 with no TWAP or deviation guard. That made the metapool LP mintable cheap while still being credited to the Strategy at close to par.

VaultYUSDT is a plain proportional-share vault: shares mint at honest NAV on deposit, then the attacker inflates NAV via stakeFor, then withdrawAll() redeems at the inflated price.

Real attack, reconstructed from the trace since multiple alert sources gave conflicting or oversimplified shapes: a Morpho USDT flash loan funds two separate legs that together mint 691,647 metapool LP tokens. Leg one runs USDT through 3pool, into the LUSD metapool, through a yVault, then a Balancer swap (capped by the pool's 50% MAX_IN_RATIO, which is why this leg alone only reaches about 230K LP). Leg two buys the depegged USDP directly and zaps it into add_liquidity on the metapool for the remaining ~461K LP. The combined LP gets staked via stakeFor(Strategy, amount), which inflates the Strategy's tracked NAV, and the attacker then calls VaultYUSDT.withdrawAll() to redeem at the inflated price, pulling real aUSDT/USDT out of the Strategy and repaying the flash loan.

The Strategy address is a proxy; deposited() and withdraw() delegatecall to the impl contract, confirmed to be the same logical contract rather than a separate one. A Uniswap V4 unlock call appears in the same tx but is incidental plumbing for the USDP purchase, not part of the value-extraction mechanism, and the PoC reproduces the full loss without it.

Reconstructed as a named exploit contract making real typed calls end to end: the flash loan callback, the real Curve/Balancer/yVault swap routes for both LP-acquisition legs, the real Convex stakeFor call, and the real VaultYUSDT.withdrawAll() redeem. No bytecode blob, no raw calldata replay, no settled-state shortcut. The attacker EOA starts at 0 USDT; every token comes from the flash loan and real market operations.

Reproduced: 344,751 USDT net profit (reported ~$345.9K, 99.7% match), 693,480e18 LP staked (real 691,647, 100.3% match), and gross Strategy loss of 597,352 USDT (reported ~$595K, matches).

Fork at parent block 25990442, exploit tx 0x5ff8150482f5473bff16b4a142a98a7f72b159df5e9dd38afd90470551640d37.


SunWeb3Sec:

- ✅ PoC authenticity: genuine on-fork reconstruction (flash loan → honest-NAV deposit → real Curve/Balancer/yVault swaps to mint metapool LP → permissionless `stakeFor` NAV inflation → `withdrawAll`). No bytecode/calldata replay, no settled-state shortcut.
- ✅ `forge test --contracts src/test/2026-09/Flamincome_exp.sol -vvv` — PASS. Net profit 344,751.209366 USDT (reported ~$345.9K), gross Strategy loss 597,352.101480 (reported ~$595K).
- ✅ README: incident count 860→861, new entry correctly hyperlinked and ordered.

Thanks for the contribution!

PR #1267, opened 2026-09-19
Add PrimeFinance PRFI oracle exploit PoC - permissionless stale signed report overwrite drains lending pool

BhariGowda:
DataStreamConsumer.verifyReport(bytes) is permissionless. It forwards to the Chainlink Verifier proxy, which checks the DON signature on the report is valid, but never checks the report is fresh. A validly-signed report with an expiresAt of 2025-11-01 was accepted and applied at a 2026-09-16 block, well past its own expiry. Stale-but-signed is accepted the same as fresh.

Confirmed permissionless three ways: calling verifyReport from an arbitrary address with no role and no prior state returns success at the parent block, the exploit block, and current head. The consumer contract is Ownable, but owner()/transferOwnership only gate config setters, not verifyReport itself. And the real attacker's own freshly-deployed contract, holding no special role, called it directly in the exploit tx.

Real attack, single tx: flash-loan 100 WHYPE from Morpho Blue, call verifyReport twice with the two real DON-signed reports captured from the exploit tx's own calldata, which reprices PRFI from about $0.0021 to about $0.1101 (roughly 52x). Buy PRFI off the thin WHYPE/PRFI pool at the now-inflated valuation, deposit it as collateral, borrow WHYPE against the inflated collateral value, repay the flash loan.

Reconstructed as a named exploit contract making real typed calls, using the real captured DON-signed report bytes for verifyReport rather than faking any storage writes. No bytecode blob, no raw calldata replay, no settled-state shortcut.

Reproduced: 425.521620381705945356 WHYPE drained from the reserve, exact match to the reported ~425.5 WHYPE. Net profit in the PoC comes to about 397.02 WHYPE against a reported ~398.70, a small gap explained by HyperswapPair's non-constant-product accounting on the actual PRFI buy (about 26.82 WHYPE moved on-chain for 795,587 PRFI, versus about 28.5 WHYPE a plain constant-product model would need for the same collateral) rather than any imprecision in the exploit mechanism itself, since the collateral and borrow amounts were sized to the real on-chain figures and the reserve drain matches to the wei.

Two things worth flagging for review: this fork requires an archival RPC (the public rpc.hyperliquid.xyz/evm only serves current storage for historical blocks, which breaks Aave's index math at a past block with a MathError). The test uses hyperliquid.drpc.org as a hardcoded fork URL inside the file rather than touching foundry.toml, since no hyperevm alias exists there yet. Also, no hyperevm alias was added to foundry.toml at all, since that's a shared-config change outside what I'd want bundled into a PoC PR.

Exploit tx: https://hyperevmscan.io/tx/0xff990876d863a61732779c341991215856c89420b84daaf31eece7ecd5ff4243
Attacker: 0x19bc1c7fd4aa93f540498499b8f5b4fc3dde5a52


SunWeb3Sec:

- ✅ PoC authenticity: genuine on-fork reconstruction. The permissionless `DataStreamConsumer.verifyReport` call is exercised directly on-fork with the real captured DON-signed reports (used as call arguments, not deployed as bytecode or replayed as raw calldata), followed by real swap/deposit/borrow calls. No bytecode blob, no calldata replay, no settled-state shortcut.
- ✅ `forge test --contracts src/test/2026-09/PrimeFinance_exp.sol -vvv` — PASS (after a transient RPC 408/rate-limit on first attempt, confirmed infra-only and retried clean). Drained 425.521620381705945356 WHYPE (exact match to reported ~425.5 WHYPE), net profit 397.02 WHYPE (reported ~398.70, small gap explained in the PoC comments by AMM accounting).
- ✅ Fork target: hardcoded `https://hyperliquid.drpc.org` — drpc.org is the same public-RPC provider already trusted for several chains in this repo's own `foundry.toml` (polygon, linea, sei, cronos), so this is treated as a recognized public RPC rather than an unfamiliar one. No `hyperevm` alias exists yet in `foundry.toml`; the author correctly avoided touching shared config.
- ✅ README: count correctly bumped, entry hyperlinked with a resolving anchor.
- ℹ️ This PR's README edit conflicted with #1266 (merged just before it, same incident date). Rebased `feat/prime` onto updated `main`, reapplied the count bump (861→862) and merged both dated entries in the incident list, and pushed the rebase back to this PR's branch before merging — no change to the PoC itself.

Thanks for the contribution!

PR #1268, opened 2026-09-19
Add Likwid margin position exploit PoC - missing pairDelta update in leverage=0 borrow freezes AMM quote

BhariGowda:
LikwidMarginPosition._executeAddCollateralAndBorrow, the leverage=0 branch of addMargin, sets delta.lendDelta, delta.mirrorDelta, and delta.marginDelta, but never sets delta.pairDelta. That field is the only one that feeds back into the AMM's pairReserves through the unlock callback. The leverage>0 branch, _executeAddLeverage, does set it correctly a few lines away. So a leverage=0 margin borrow leaves pairReserves completely frozen, even though real value is moving through the position.

SwapMath.getAmountOut(poolState.pairReserves, ...), called at the top of the same function on every invocation, keeps quoting off those frozen reserves. Since the reserves never actually update, every leverage=0 borrow gets requoted at the exact same marginal price with zero AMM slippage, no matter how many times it's called.

addMargin is external payable, gated only by owning the position NFT the caller mints for itself in the same call. No admin role, no signer, fully unprivileged.

Real attack, single tx: the attacker's contract, already holding 218.58M of the token from before the exploit block (confirmed genuine pre-state, not sourced same-tx), called addMargin 24 times with leverage=0 and marginForOne=true. The first 15 calls each borrow an identical 4.787036566 BNB, the static-reserve signature of the bug. After that the borrow size tapers, not because the price ever moves, but because borrowMaxAmount is separately capped at 20% of the shrinking realReserves. Total: 211,832,397.21 tokens deposited across all calls, 74.310377810 BNB drained.

Reconstructed as a named exploit contract making the real typed addMargin(PoolKey, CreateParams) calls into the real LikwidMarginPosition and LikwidVault, replaying the actual 24-call sequence rather than a single shortcut call. The attacker's pre-held token balance is funded via a prank-transfer, not deal() or a minted balance, since that reflects the real pre-exploit state. No bytecode blob, no raw calldata replay, no settled-state shortcut.

Reproduced: 74.310377807370642618 BNB drained, a 3-wei difference from the real 74.310377810 BNB.

Two notes on the fork setup: the bsc alias in foundry.toml (binance.llamarpc.com) is dead and non-archival regardless, so the test reads a BSC_RPC env override pointing at an archival endpoint rather than touching foundry.toml. It also needs evm_version = cancun since the vault's V4-style unlock uses transient storage, applied as an inline forge-config annotation on the test rather than a global config change.

Run: BSC_RPC=https://bsc-mainnet.public.blastapi.io forge test --contracts src/test/2026-09/Likwid_exp.sol -vvv

Exploit tx: https://bscscan.com/tx/0x83cbd07d59aedc2f114c351c568d3386ff5bc7caa87d12bc6494e2dc7bf16f4d
Attacker: 0x90bde1e0bb16b3deeb9d638acf8d01f19fd2f31e


SunWeb3Sec:

- ✅ PoC authenticity: genuine on-fork reconstruction. The real 24-call `addMargin` sequence is replayed exactly against the deployed `LikwidMarginPosition`/`LikwidVault`, exploiting the missing `pairDelta` update directly on-fork. The attacker's pre-held TOKEN balance is confirmed genuine pre-exploit state (present since before the fork block) and is only used as working capital, not as an unreconstructed manipulation step — the actual exploit (repeated frozen-quote borrows) is fully performed in Solidity. No bytecode blob, no calldata replay, no settled-state shortcut.
- ✅ `forge test --contracts src/test/2026-09/Likwid_exp.sol -vvv` (with `BSC_RPC` pointed at an archival BSC endpoint, since the `bsc` alias in `foundry.toml` is dead/non-archival) — PASS. Drained 74.310377807370642618 BNB, matching the reported ~74.310377810 BNB to within 3 wei.
- ✅ Fork target: `https://bsc-mainnet.public.blastapi.io` is a known-good public archival BSC RPC; no `foundry.toml` changes needed since the PoC reads it via an env-var override, not a hardcoded value.
- ✅ README: count correctly bumped, entry hyperlinked with a resolving anchor.
- ℹ️ This PR's README edit conflicted with #1266/#1267 (merged just before it). Rebased `feat/likwid` onto updated `main`, reapplied the count bump (862→863) and reordered the incident-list/detail entries by date, and pushed the rebase back to this PR's branch before merging — no change to the PoC itself.

Thanks for the contribution!

PR #1269, opened 2026-09-20
Add Nimiq exploit PoC - GSN forwarder skips signature check, relying solely on attacker-chosen paymaster

BhariGowda:
ERC20PermitHTLCHandler serves as its own OpenGSN forwarder, paymaster, and recipient all at once. Its execute(ForwardRequest, bytes32, bytes32, bytes, bytes) is the GSN forwarder entrypoint, and it discards every one of its five parameters, including the signature and nonce, then blindly calls openPrivate(request.from, ...), which does token.transferFrom(request.from, handler, amount). There is no EIP-712 signature check, no nonce check, and the handler's own precondition check is skipped entirely.

The only place any of that validation happens is preRelayedCall(), a callback OpenGSN's RelayHub invokes on whichever paymaster the relay itself specifies. Since these handlers are wired as their own paymaster, and RelayHub lets any registered relay pick any paymaster and forwarder pair for a call, an attacker can register their own relay with an accept-everything paymaster and simply declare request.from to be a victim, and the handler has no way to tell the difference.

execute() does check onlyRelayHub (msg.sender == getHubAddr()), but that's not a meaningful gate here since the actual trust boundary that's broken is one level up, in RelayHub trusting whatever paymaster a registered relay names. owner() exists on the handler too, but it only guards registerToken, setRelayHub, and withdraw functions, nothing on the drain path itself. Confirmed against the real deployed source for both handler instances.

Real attack: the attacker's EIP-7702-delegated EOA staked 1 POL/MATIC and registered as a relay manager, worker, and its own accept-everything paymaster against the real RelayHub v2.2.0 and StakeManager, pointing the forwarder at both real HTLC handlers. It then submitted forged open() requests through RelayHub.relayCall() with request.from set to a swap-liquidity wallet holding unlimited approvals to both handlers, opening HTLCs for the wallet's full USDC, USDT0, and USDC.e balances. Each forged HTLC named the attacker's own precomputed CREATE2 address as recipient with hashlock sha256(0x01). A final tx deployed that CREATE2 contract and redeemed all three HTLCs with secret 0x01.

Reconstructed as a named exploit contract making real typed calls against the actual deployed RelayHub, StakeManager, and both HTLC handlers. Registration is done the real way: setRelayManagerOwner, stakeForRelayManager with 1 POL, authorizeHubByManager, addRelayWorkers, no vm.store shortcuts. The forged open() call goes through a real RelayRequest via RelayHub.relayCall() from the worker EOA, satisfying RelayHub's actual gas accounting and RelayHubValidator.verifyTransactionPacking calldata rules along the way, not a simplified version of the flow. No bytecode blob, no raw calldata replay, no settled-state shortcut.

The setup tx (staking and relay registration) is generic GSN infrastructure setup with nothing victim-specific in it, since the victim's unlimited approvals and balances predate both txs, so the PoC forks one block before the setup tx and rebuilds the registration fresh rather than replaying that tx verbatim.

Reproduced: $50,463.792096 drained across USDC, USDT0, and USDC.e, matching the pre-exploit victim balances and the reported ~$50,463 loss to the cent.

Setup tx: https://polygonscan.com/tx/0xb067efae73637f3564f58af7f6027afc497e81e47624b0048636085c678858c0
Exploit tx: https://polygonscan.com/tx/0xb2ca76dfbfe571742b4b66465b777ab1e06988a8632be1631bef9654cc64d169
Attacker: 0x2258491525C21f334c5a2dc22CE55e55023FC45D


SunWeb3Sec:

- ✅ PoC is an authentic reconstruction: real relay registration via `StakeManager`/`RelayHub`, a named `EvilPaymaster` contract, forged `relayCall()` through the actual OpenGSN infra — no bytecode/calldata replay, no settled-state shortcut.
- ✅ `forge test --contracts src/test/2026-09/Nimiq_exp.sol -vvv` — PASS, total drained $50,463.792096, matching the reported ~$50,463 loss.
- ✅ README: incident count 863 → 864, new TOC entry correctly ordered and hyperlinked with a resolving anchor, blank-line separation intact.

Thanks for the contribution!

PR #1270, opened 2026-09-20
Add PoC: Startale ERC-7579 transient-storage re-init drain

BhariGowda:
## Root cause

StartaleSmartAccount uses EIP-1153 transient storage as its account-initialization gate. AccountProxy's constructor calls setInitializable() (tstore(INIT_SLOT, 1)) before delegatecalling into the legit initializeAccount(bytes) to install the owner's validator. _initializeAccount has no persistent "already initialized" guard - the transient flag is the only check (requireInitializable() = tload(INIT_SLOT)).

Transient storage is scoped to the whole transaction, not the constructor's call frame. So after StartaleAccountFactory.createAccount() deploys and legitimately initializes an account, INIT_SLOT stays 1 on that account for the rest of the transaction. Any external, unrelated contract can call initializeAccount(evilInitData) on the freshly-deployed account again in the same tx, pass the same transient check, and delegatecall into an attacker-supplied "bootstrap" that sweeps the account's pre-funded balance. No signature, no privileged caller, zero attacker capital.

Confirmed against verified deployed source for StartaleSmartAccount (0x000000b8f5f723a680d3d7ee624fe0bc84a6e05a) and StartaleAccountFactory (0x0000003b3e7b530b4f981ae80d9350392defef90), and empirically in the PoC: the second initializeAccount call succeeds and drains the account. If the transient flag were constructor-scoped, this would revert and the test would fail.

## Example tx

https://etherscan.io/tx/0x1a021a27c8db5ecd6428a567cedf008872f205c55e590d4dcd2ea10e31a036a2 (block 25987506) - 427.551160 USDC drained across 23 pre-funded accounts in this batch alone. This is 1 of 41 known batches; reported aggregate loss is ~$2,876 USDC+USDT across ~330 victim accounts.

## PoC

Reconstructed as a named exploit contract plus a minimal malicious bootstrap, both making real typed calls into the actual factory and account functions (no bytecode/calldata replay). Reproduces 3 of 23 victim accounts from the example tx, forked at the parent block (25987505). Assert: swept total = 32.555561 USDC, matching the proportional share of the 3 reconstructed accounts.

forge test --contracts src/test/2026-09/Startale_exp.sol -vvv


SunWeb3Sec:

- ✅ PoC is an authentic reconstruction: forks at the parent block (before the victim accounts exist), then reconstructs both the legitimate `factory.createAccount()` init and the same-tx malicious `initializeAccount()` re-init that abuses the transient-storage flag — no bytecode/calldata replay, no settled-state shortcut.
- ✅ `forge test --contracts src/test/2026-09/Startale_exp.sol -vvv` — PASS, 32.555561 USDC swept from the 3 reconstructed victim accounts, matching the PR's claimed figure.
- ⚠️ This branch was opened before PR #1269 merged, so it still bumped the incident count 863→864 — which collided with #1269's identical bump and would have silently left the count wrong (864 instead of 865) on a plain merge. I rebased the branch onto the post-#1269 `main` and pushed a follow-up commit correcting the count to 865; TOC/detail entries and anchor are correctly ordered and linked.

Thanks for the contribution!

